PR #15, opened 2019-11-12
Adopt official mattn/goveralls integration

jbrockopp:
This follows the officially documented [goveralls](https://github.com/mattn/goveralls) integration with [Actions](https://github.com/features/actions):

https://github.com/mattn/goveralls#github-actions

Note that we are required to use the Github token instead of the Coveralls token:

https://github.com/mattn/goveralls/pull/144#discussion_r342997287

Might also help with fork PR runs?

https://github.com/mattn/goveralls/pull/144#discussion_r343013930

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,9 +18,11 @@ jobs:
 
     - name: test
       run: |
-        go test -covermode=count -coverprofile=coverage.out ./...
+        go test -race -covermode=atomic -coverprofile=coverage.out ./...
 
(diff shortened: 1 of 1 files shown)


jbrockopp:
@wass3r I've tried rerunning it and it still fails with a `500` error from Coveralls 😢 

I'm thinking something is stuck on their end 😬 

The tests themself pass and the push event uploaded to Coveralls.

https://coveralls.io/jobs/55553825

PR #21, opened 2019-11-19
Add the ability to route builds

kneal:


cmd/server/runtime.go
```diff
@@ -19,6 +19,8 @@ import (
 // helper function to setup the runtime from the CLI arguments.
 func setupRuntime(c *cli.Context) (runtime.Engine, error) {
 	logrus.Debug("Creating runtime client from CLI configuration")
+
+	c.Set("queue-route-runtime", c.String("runtime-driver"))
```
golangcibot: Error return value of `c.Set` is not checked (from `errcheck`)

PR #26, opened 2019-11-26
Allow only server to make API calls with secret

kneal:


router/middleware/perm/perm.go
```diff
@@ -0,0 +1,31 @@
...
+func MustServer() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		u := user.Retrieve(c)
+
+		if strings.EqualFold(u.GetName(), "vela-server") {
+			return
+		}
+
+		msg := fmt.Sprintf("User %s is not a platform admin", u.GetName())
+		c.Error(fmt.Errorf(msg))
```
golangcibot: Error return value of `c.Error` is not checked (from `errcheck`)

PR #28, opened 2019-11-29
Update client maps to be sync map

kneal:


executor/linux/service.go
```diff
@@ -81,7 +82,12 @@ func (c *client) PlanService(ctx context.Context, ctn *pipeline.Container) error
 func (c *client) ExecService(ctx context.Context, ctn *pipeline.Container) error {
 	b := c.build
 	r := c.repo
-	l := c.serviceLogs[ctn.ID]
+
+	result, ok := c.serviceLogs.Load(ctn.ID)
+	if !ok {
+		return fmt.Errorf("unable to get log from client")
```
jbrockopp: Thoughts on updating the error message for the resource here?

Maybe something like:

`unable to get service log from client`

executor/linux/step.go
```diff
@@ -139,7 +139,12 @@ func (c *client) PlanStep(ctx context.Context, ctn *pipeline.Container) error {
 func (c *client) ExecStep(ctx context.Context, ctn *pipeline.Container) error {
 	b := c.build
 	r := c.repo
-	l := c.stepLogs[ctn.ID]
+
+	result, ok := c.stepLogs.Load(ctn.ID)
+	if !ok {
+		return fmt.Errorf("unable to get log from client")
```
jbrockopp: Potentially something similar to above?

i.e. `unable to get step log from client`

executor/linux/service.go
kneal: Yeah being a bit more verbose would definitely help the parsing when you're looking at the output in a log aggregator. Will update!

executor/linux/step.go
kneal: Same as above. 😄

PR #51, opened 2020-01-03
fix: unable to pull shared secrets

kneal:


executor/linux/secret.go
```diff
@@ -165,8 +165,7 @@ func (c *client) getShared(s *pipeline.Secret) (*library.Secret, error) {
 
 	// variables necessary for secret
 	var team string
-
-	org := c.repo.GetOrg()
+	var org string
```
golangcibot: declarations should never be cuddled (from `wsl`)

executor/linux/secret.go
```diff
@@ -165,8 +165,8 @@ func (c *client) getShared(s *pipeline.Secret) (*library.Secret, error) {
...
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

PR #99, opened 2020-07-07
feat(api): add open api (swagger) spec

meinenec:
closes https://github.com/go-vela/community/issues/51

using [go-swagger](https://github.com/go-swagger/go-swagger) to generate
install following [their docs](https://goswagger.io/install.html#installing)

build with `swagger generate spec -m -o api-spec/vela-worker-v0.4.3.json -x "github.com/docker/docker"`
validate with `swagger validate api-spec/vela-worker-v0.4.3.json`


meinenec:
The golangci complaints are not related to my change. Would be good to address them as they come up, but shouldn't be a part of this PR.

PR #110, opened 2020-08-19
chore: set hostname in docker-compose

meinenec:
Adding hostnames to worker containers in docker-compose to allow the server to address the worker api locally.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@ services:
       context: .
     container_name: worker_1
     image: worker:local
+    hostname: http://worker_1
     networks:
       - vela
     environment:
@@ -33,6 +34,7 @@ services:
       context: .    
     container_name: worker_2
     image: worker:local
+    hostname: http://worker_2
     networks:
       - vela
     environment:


docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
       context: .
     container_name: worker_1
     image: worker:local
+    hostname: http://worker_1
```
jbrockopp, line 13: Should the hostname have the protocol (http://) in it?

meinenec: I'm not sure :) I'm okay to remove it but then I think we'll need another environment variable to state the protocol a worker is available on. I am okay to make that change if you'd like.

jbrockopp: It was more my curiosity than anything 🤷 

This `hostname` variable is a Docker compose attribute (not set directly in the worker) so I think I'm ok with it 👍 

I'd have larger concerns if we were directly setting this in the worker application.

kneal: Actually, since you brought that up we do have flag that might be more appropriate:
https://github.com/go-vela/worker/blob/master/cmd/vela-worker/flags.go#L70-L74

Essentially it's for setting the Addr of the worker which I imagine is a routable address

jbrockopp: @kneal 

FYI: That's not setting the addr of the worker.

That's setting the addr for the Vela server that the worker communicates with 👍 

For example, this is what that is [set to in our compose stack](https://github.com/go-vela/worker/blob/master/docker-compose.yml#L24):

```
VELA_SERVER_ADDR: http://vela:8080
```

meinenec: That's a good point @kneal 

Thanks for the feedback and merging on this. I'll look into and potentially make a future PR to use `WORKER_SERVER_ADDR`

PR #113, opened 2020-09-03
feat: allow server to run with TLS

meinenec:
Allows (optionally) setting a certificate which is then used in [RunTLS](https://pkg.go.dev/github.com/gin-gonic/gin?tab=doc#Engine.RunTLS)

If certs are not provided, the functionality stays the same as the current state.


wass3r (approved):
lgtm

jbrockopp (approved):
LGTM

PR #116, opened 2020-09-28
fix: revert PR 113

meinenec:
Removes the ability to run with TLS.

Will re-add after release of 0.6.0.

### DIFF
--- a/cmd/vela-worker/flags.go
+++ b/cmd/vela-worker/flags.go
@@ -77,16 +77,6 @@ func flags() []cli.Flag {
 			Name:    "server.secret",
 			Usage:   "secret used for server <-> worker communication",
 		},
-		&cli.StringFlag{
-			EnvVars: []string{"WORKER_SERVER_CERT", "VELA_SERVER_CERT", "SERVER_CERT"},
-			Name:    "server.cert",
-			Usage:   "optional TLS certificate for https",
-		},
-		&cli.StringFlag{
-			EnvVars: []string{"WORKER_SERVER_CERT_KEY", "VELA_SERVER_CERT_KEY", "SERVER_CERT_KEY"},
-			Name:    "server.cert-key",
-			Usage:   "optional TLS certificate key",
-		},
 	}
 
 	// Executor Flags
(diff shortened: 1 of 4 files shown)


KellyMerrick (approved):
lgtm

wass3r (approved):

jbrockopp (changes requested):
Do we need to do this?

meinenec:
@jbrockopp I discovered a bug with the current implementation. We had:
```go
if len(w.Config.Certificate.Cert) > 0 {	
	return _server.RunTLS(w.Config.API.Port, w.Config.Certificate.Cert, w.Config.Certificate.Key)	
}
```

where the value of `w.Config.Certificate.Cert` is a string containing the path to the cert. If the path is set, but doesn't exist, the worker errors and fails to start.

The choices were either to either revert or fix the issue. The fix _should_ be a simple check to see if the files exist, but I opted to revert until after the release to avoid any other unknown issues.

jbrockopp:
@meinenec Hmm... I might be missing something 👍 

Could you elaborate on why the behavior you describe is a bug?

meinenec:
Currently, it fails with the error (taken from a local run):
```
[GIN-debug] Listening and serving HTTPS on :8080
{"level":"trace","msg":"serving traffic on :8080","time":"2020-09-28T14:10:45Z"}
{"level":"trace","msg":"Creating redis queue client from CLI configuration","time":"2020-09-28T14:10:45Z"}
[GIN-debug] [ERROR] open /foo: no such file or directory
{"level":"error","msg":"failing worker server: open /foo: no such file or directory","time":"2020-09-28T14:10:45Z"}
{"level":"fatal","msg":"shutting down worker","time":"2020-09-28T14:10:45Z"}
```

I personally don't find the error message to be very meaningful. I understand it can't open a file, but it's not clear why it's opening the file. In practice, this would probably be remediated by the filepath being more recognizable.

I understand that this isn't the most impactful bug, but like I pointed out I want to revert the change to avoid this and other unknown or unexpected behaviors.

PR #119, opened 2020-10-16
feat: use new version pkg

jbrockopp:
Dependent on https://github.com/go-vela/types/pull/112

Updating to the new `go-vela/types/version` package to improve the information provided surrounding the application.

If I compile the binary specifically for the OS on my workstation, I'm met with this from the `--version` flag:

```sh
$ release/vela-worker --version
vela-worker version v0.6.0
```

When I run `make up` locally, here is the output from http://localhost:8080/version:

```json
{
  "canonical": "v0.6.0",
  "major": 0,
  "minor": 6,
  "patch": 0,
  "metadata": {
    "architecture": "amd64",
    "build_date": "2020-10-16T09:59:31Z",
    "compiler": "gc",
    "git_commit": "da574fdb6c6ff93cc17349d02a9b17394108dc59",
    "go_version": "go1.15.2",
    "operating_system": "linux"
  }
}
```

**NOTE: This also updates the GitHub Actions pipelines to use our `make` commands for building the binary.**


router/middleware/header.go
```diff
@@ -57,26 +57,24 @@ func Secure(c *gin.Context) {
...
-	apiVersion := version.Version
+	v := version.New()
 
 	if gin.Mode() == "debug" {
-		c.Header("X-Vela-Version", apiVersion.String())
+		c.Header("X-Vela-Version", v.Semantic())
 	} else { // in prod we don't want the build number metadata
-		apiVersion.Metadata = ""
-		c.Header("X-Vela-Version", apiVersion.String())
+		c.Header("X-Vela-Version", v.Semantic())
```
wass3r, line 78: maybe i missed something, but do we need this if statement if we're returning the same thing?

wass3r (changes requested):

router/middleware/header.go
jbrockopp: No we don't need the if statement 👍 

If you'd like I can comment out the `if` but I left it in favor of being explicit on how we could customize it in the future.

In short, my thought process was by leaving it there, if (or when) we want to look at customizing how that header is populated from the API when running in `debug` mode then the shell of the code is there.

wass3r (approved):

router/middleware/header.go
wass3r: K, not a big deal to leave it in 👍

kneal (approved):
LGTM 🐬

PR #122, opened 2020-10-26
docs: update info for local development

jbrockopp:
Related to https://github.com/go-vela/server/pull/202 and https://github.com/go-vela/server/pull/204

It's been a long time since the local development docs have been updated so thought I'd give it a try.

For convenience, here are links to the old/new local development docs:

* [Old](https://github.com/go-vela/worker/blob/master/.github/DOCS.md)
* [New](https://github.com/go-vela/worker/blob/docs/local/DOCS.md)

Found within this PR are the following changes:

* moved the `DOCS.md` to the root of the repository so it's easier to find (previously was nested under `.github`)
* updated the `DOCS.md` content to be more focused on starting the Vela stack using `make`
* updated the `DOCS.md` with information on the different compose services
* added to the `DOCS.md`  instructions for running builds locally when the Vela stack is started
* updated the `CONTRIBUTING.md` to reference the `DOCS.md` to reduce duplication of docs
* updated the `docker-compose.yml` to use `'` for strings (previously, we were using `'` and `"` interchangeably)
* updated the `.gitignore` to exclude the `secrets.env` file needed for [go-vela/server](https://github.com/go-vela/server)
* added a `secrets.env.example` file so users can reference it as a guide to creating their own


cmd/vela-worker/main.go
```diff
@@ -39,8 +39,8 @@ func main() {
 	// Worker Information
 
 	app.Name = "vela-worker"
-	app.HelpName = "vela-executor"
-	app.Usage = "Vela executor package for integrating with different executors"
+	app.HelpName = "vela-worker"
+	app.Usage = "Vela build daemon designed for executing pipelines"
```
wass3r, line 43: nice catches :)

wass3r (approved):

wass3r (approved):

KellyMerrick (approved):
lgtm

PR #124, opened 2020-10-30
feat: phone home to server on time interval

meinenec:
This PR will register a worker with the server on start (or update itself if it currently exists) and then continue to check in with the server on the specified interval.

### DIFF
--- a/cmd/vela-worker/flags.go
+++ b/cmd/vela-worker/flags.go
@@ -26,7 +26,7 @@ func flags() []cli.Flag {
 			Usage:   "set hostname for the worker",
 		},
 
-		&cli.Int64Flag{
+		&cli.DurationFlag{
 			EnvVars: []string{"VELA_WORKER_CHECK_IN", "VELA_CHECK_IN", "CHECK_IN"},
 			Name:    "checkIn",
 			Usage:   "time to wait in between checking in with the server",
--- a/cmd/vela-worker/flags.go
+++ b/cmd/vela-worker/flags.go
@@ -30,7 +30,7 @@ func flags() []cli.Flag {
 			EnvVars: []string{"WORKER_CHECK_IN", "VELA_CHECK_IN", "CHECK_IN"},
 			Name:    "checkIn",
 			Usage:   "time to wait in between checking in with the server",
-			Value:   15,
+			Value:   15 * time.Minute,
 		},
 
 		// API Flags
--- a/cmd/vela-worker/flags.go
+++ b/cmd/vela-worker/flags.go
@@ -27,7 +27,7 @@ func flags() []cli.Flag {
 		},
 
 		&cli.DurationFlag{
-			EnvVars: []string{"VELA_WORKER_CHECK_IN", "VELA_CHECK_IN", "CHECK_IN"},
+			EnvVars: []string{"WORKER_CHECK_IN", "VELA_CHECK_IN", "CHECK_IN"},
 			Name:    "checkIn",
 			Usage:   "time to wait in between checking in with the server",
 			Value:   15,
--- a/cmd/vela-worker/flags.go
+++ b/cmd/vela-worker/flags.go
@@ -26,6 +26,13 @@ func flags() []cli.Flag {
 			Usage:   "set hostname for the worker",
 		},
 
+		&cli.Int64Flag{
+			EnvVars: []string{"VELA_WORKER_CHECK_IN", "VELA_CHECK_IN", "CHECK_IN"},
+			Name:    "checkIn",
+			Usage:   "time to wait in between checking in with the server",
+			Value:   15,
+		},
+
 		// API Flags
 
 		&cli.StringFlag{
--- a/cmd/vela-worker/operate.go
+++ b/cmd/vela-worker/operate.go
@@ -34,9 +34,9 @@ func (w *Worker) operate() error {
 	registryWorker.SetRoutes(w.Config.Queue.Routes)
 	registryWorker.SetActive(true)
 	registryWorker.SetLastCheckedIn(time.Now().UTC().Unix())
-	w.register(registryWorker)
+	err = w.register(registryWorker)
 	if err != nil {
-		logrus.Error("unable to register worker with the server")
+		logrus.Error(err)
 	}
 
 	// spawn goroutine for phoning home
@@ -51,7 +51,7 @@ func (w *Worker) operate() error {
 
 			// if unable to update the worker, log the error but allow the worker to continue running
 			if err != nil {
-				logrus.Error("unable to update worker in database: %w", err)
+				logrus.Errorf("unable to update worker %s in database: %v", registryWorker.GetHostname(), err)
 			}
 		}
 	}()
--- a/cmd/vela-worker/operate.go
+++ b/cmd/vela-worker/operate.go
@@ -46,7 +46,7 @@ func (w *Worker) operate() error {
 	go func() {
 		for {
 			// sleep for the configured time
-			time.Sleep(time.Minute * time.Duration(*w.Config.CheckIn))
+			time.Sleep(*w.Config.CheckIn)
 
 			// set checking time to now and call the server
 			registryWorker.SetLastCheckedIn(time.Now().Unix())
--- a/cmd/vela-worker/operate.go
+++ b/cmd/vela-worker/operate.go
@@ -5,7 +5,10 @@
 package main
 
 import (
+	"time"
+
 	"github.com/go-vela/pkg-queue/queue"
+	"github.com/go-vela/types/library"
 
 	"github.com/sirupsen/logrus"
 
@@ -23,6 +26,35 @@ func (w *Worker) operate() error {
 	if err != nil {
 		return err
 	}
+	// Define the database representation of the worker
+	// and register itself in the database
+	workerActive := true
+	workerLastCheckedIn := time.Now().Unix()
+	registryWorker := library.Worker{
+		Hostname:      &w.Config.Hostname,
+		Address:       &w.Config.Server.Address,
+		Routes:        &w.Config.Queue.Routes,
+		Active:        &workerActive,
+		LastCheckedIn: &workerLastCheckedIn,
+	}
+	w.register(&registryWorker)
+
+	// spawn goroutine for phoning home
+	go func() {
+		for {
+			// sleep for the configured time
+			time.Sleep(time.Minute * time.Duration(*w.Config.CheckIn))
+
+			// set checking time to now and call the server
+			registryWorker.SetLastCheckedIn(time.Now().Unix())
+			_, _, err := w.VelaClient.Worker.Update(registryWorker.GetHostname(), &registryWorker)
+
+			// if unable to update the worker, log the error but allow the worker to continue running
+			if err != nil {
+				logrus.Error("unable to update worker in database: %w", err)
+			}
+		}
+	}()
 
 	// setup the queue
 	//
--- a/cmd/vela-worker/operate.go
+++ b/cmd/vela-worker/operate.go
@@ -38,6 +38,9 @@ func (w *Worker) operate() error {
 		LastCheckedIn: &workerLastCheckedIn,
 	}
 	w.register(&registryWorker)
+	if err != nil {
+		logrus.Error("unable to register worker with the server")
+	}
 
 	// spawn goroutine for phoning home
 	go func() {
--- a/cmd/vela-worker/operate.go
+++ b/cmd/vela-worker/operate.go
@@ -43,7 +43,7 @@ func (w *Worker) operate() error {
 	go func() {
 		for {
 			// sleep for the configured time
-			time.Sleep(*w.Config.CheckIn)
+			time.Sleep(w.Config.CheckIn)
 
 			// set checking time to now and call the server
 			registryWorker.SetLastCheckedIn(time.Now().UTC().Unix())
--- a/cmd/vela-worker/operate.go
+++ b/cmd/vela-worker/operate.go
@@ -28,16 +28,13 @@ func (w *Worker) operate() error {
 	}
 	// Define the database representation of the worker
 	// and register itself in the database
-	workerActive := true
-	workerLastCheckedIn := time.Now().Unix()
-	registryWorker := library.Worker{
-		Hostname:      &w.Config.Hostname,
-		Address:       &w.Config.Server.Address,
-		Routes:        &w.Config.Queue.Routes,
-		Active:        &workerActive,
-		LastCheckedIn: &workerLastCheckedIn,
-	}
-	w.register(&registryWorker)
+	registryWorker := new(library.Worker)
+	registryWorker.SetHostname(w.Config.Hostname)
+	registryWorker.SetAddress(w.Config.Server.Address)
+	registryWorker.SetRoutes(w.Config.Queue.Routes)
+	registryWorker.SetActive(true)
+	registryWorker.SetLastCheckedIn(time.Now().UTC().Unix())
+	w.register(registryWorker)
 	if err != nil {
 		logrus.Error("unable to register worker with the server")
 	}
@@ -49,8 +46,8 @@ func (w *Worker) operate() error {
 			time.Sleep(*w.Config.CheckIn)
 
 			// set checking time to now and call the server
-			registryWorker.SetLastCheckedIn(time.Now().Unix())
-			_, _, err := w.VelaClient.Worker.Update(registryWorker.GetHostname(), &registryWorker)
+			registryWorker.SetLastCheckedIn(time.Now().UTC().Unix())
+			_, _, err := w.VelaClient.Worker.Update(registryWorker.GetHostname(), registryWorker)
 
 			// if unable to update the worker, log the error but allow the worker to continue running
 			if err != nil {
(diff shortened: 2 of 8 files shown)


cmd/vela-worker/flags.go
```diff
@@ -26,6 +26,13 @@ func flags() []cli.Flag {
 			Usage:   "set hostname for the worker",
 		},
 
+		&cli.Int64Flag{
```
JordanSussman, line 29: Why not utilize the [cli.DurationFlag](https://github.com/urfave/cli/blob/master/flag_duration.go#L10)?

meinenec: This is just being used in time.Sleep so either should work fine. I wasn't sure that more granular control was useful so I never updated this after my initial implementation. I think it would be fine to use either Int64 or DurationFlag if there is any preference.

kneal: I think duration would be more ideal since it's more user friendly to the administrator.

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// register is a helper function to register
+// the worker in the database, updating the item
+// if the worker already exists
+func (w *Worker) register(config *library.Worker) error {
+	// check to see if the worker already exists in the database
+	_, _, err := w.VelaClient.Worker.Get(config.GetHostname())
+	if err != nil {
```
kneal, line 18: I wonder if it's worth capturing `resp` and doing an additional check to ensure the error that's returned is a `404` to know whether or not to make the call. Right now, your server response could be any error code 4XX or 5XX and you send a call.

meinenec: I think that's a really good idea. I'll add that!

meinenec: Thanks for the suggestion. I also updated the errors in that function to log instead of returning the error. For the time being I don't think we want this to block to workers from being deployed.

cmd/vela-worker/flags.go
```diff
@@ -26,6 +26,13 @@ func flags() []cli.Flag {
 			Usage:   "set hostname for the worker",
 		},
 
+		&cli.DurationFlag{
+			EnvVars: []string{"VELA_WORKER_CHECK_IN", "VELA_CHECK_IN", "CHECK_IN"},
```
jbrockopp, line 30: For consistency with the other env vars, I think we want this to be `WORKER_CHECK_IN` 👍 

```suggestion
			EnvVars: []string{"WORKER_CHECK_IN", "VELA_CHECK_IN", "CHECK_IN"},
```

cmd/vela-worker/operate.go
```diff
@@ -23,6 +26,38 @@ func (w *Worker) operate() error {
...
+	// and register itself in the database
+	workerActive := true
+	workerLastCheckedIn := time.Now().Unix()
+	registryWorker := library.Worker{
+		Hostname:      &w.Config.Hostname,
+		Address:       &w.Config.Server.Address,
+		Routes:        &w.Config.Queue.Routes,
+		Active:        &workerActive,
+		LastCheckedIn: &workerLastCheckedIn,
+	}
```
jbrockopp, line 39: Thoughts on using our setters here?

This gives us less LOC and more closely follows how we write code for other parts of Vela 👍 

Also, we'll want to make sure that we're using the `UTC` format for the timestamps.

```suggestion
	registryWorker := new(library.Worker)
	registryWorker.SetHostname(&w.Config.Hostname)
	registryWorker.SetAddress(&w.Config.Server.Address)
	registryWorker.SetRoutes(&w.Config.Queue.Routes)
	registryWorker.SetActive(true)
	registryWorker.SetLastCheckedIn(time.Now().UTC().Unix())
```

cmd/vela-worker/operate.go
```diff
@@ -23,6 +26,38 @@ func (w *Worker) operate() error {
...
+
+	// spawn goroutine for phoning home
+	go func() {
+		for {
+			// sleep for the configured time
+			time.Sleep(*w.Config.CheckIn)
+
+			// set checking time to now and call the server
+			registryWorker.SetLastCheckedIn(time.Now().Unix())
```
jbrockopp, line 52: We'll want to make sure that we're using the `UTC` format for the timestamps 👍 

```suggestion
			registryWorker.SetLastCheckedIn(time.Now().UTC().Unix())
```

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,44 @@
...
+
+// register is a helper function to register
+// the worker in the database, updating the item
+// if the worker already exists
+func (w *Worker) register(config *library.Worker) {
+	// check to see if the worker already exists in the database
+	_, resp, err := w.VelaClient.Worker.Get(config.GetHostname())
+	if resp.StatusCode == http.StatusNotFound {
+		// worker does not exist, create it
+		logrus.Info("registering worker with the server")
```
jbrockopp, line 22: Could we augment this log message to provide a bit more information about the worker being registered?

```suggestion
		logrus.Infof("registering worker %s with the server", config.GetHostname())
```

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,44 @@
...
+		_, _, err := w.VelaClient.Worker.Add(config)
+		if err != nil {
+			// log the error instead of returning so the operation doesn't block worker deployment
+			logrus.Error("unable to register worker with the server: %w", err)
```
jbrockopp, line 26: Same thing here 👍 

```suggestion
			logrus.Error("unable to register worker %s with the server: %w", config.GetHostname(), err)
```

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+		return
+	}
+
+	// if there was an error other than a 404, log the error.
+	if err != nil {
+		// log the error instead of returning so the operation doesn't block worker deployment
+		logrus.Error("unable to get worker from server: %w", err)
```
jbrockopp, line 34: Same thing here 👍 

```suggestion
		logrus.Error("unable to get worker %s from server: %w", config.GetHostname(), err)
```

cmd/vela-worker/worker.go
```diff
@@ -41,6 +41,7 @@ type (
 	Config struct {
 		API      *API
 		Build    *Build
+		CheckIn  *time.Duration
```
jbrockopp, line 44: Wonder if we should consider not making this a pointer?

Similar to the `Hostname` field, then you wouldn't have to dereference this anytime you use the value 👍

cmd/vela-worker/run.go
```diff
@@ -80,6 +83,8 @@ func run(c *cli.Context) error {
 				Limit:   c.Int("build.limit"),
 				Timeout: c.Duration("build.timeout"),
 			},
+			// build configuration
+			CheckIn: &checkIn,
```
jbrockopp, line 87: If we make `CheckIn` not be a pointer, then we can modify this line:

```suggestion
			CheckIn: c.Duration("checkIn"),
```

jbrockopp (changes requested):

cmd/vela-worker/operate.go
meinenec: I've been thinking on this a little and I agree that consistency is best.

cmd/vela-worker/register.go
meinenec: I was thinking of logging more here but I was assuming hostname is something that we already know. I was thinking of logging the client info such as server address or the full payload that's being uploaded.

I suppose adding the hostname at least ensures that the config is populated. I'll add that as a starting point.

cmd/vela-worker/operate.go
meinenec: Good catch on using UTC as well 😄

cmd/vela-worker/operate.go
```diff
@@ -23,6 +26,35 @@ func (w *Worker) operate() error {
...
+	registryWorker := new(library.Worker)
+	registryWorker.SetHostname(w.Config.Hostname)
+	registryWorker.SetAddress(w.Config.Server.Address)
+	registryWorker.SetRoutes(w.Config.Queue.Routes)
+	registryWorker.SetActive(true)
+	registryWorker.SetLastCheckedIn(time.Now().UTC().Unix())
+	w.register(registryWorker)
+	if err != nil {
+		logrus.Error("unable to register worker with the server")
+	}
```
jbrockopp, line 40: I think we should consider having the `register()` function return an `error` 👍 

I think we could then update this code to have `logrus` output that error.

That way, we can create a good error message in the `register()` function and have it output here 👍 

```suggestion
	w.register(registryWorker)
	if err != nil {
		logrus.Error(err)
	}
```

cmd/vela-worker/operate.go
```diff
@@ -23,6 +26,35 @@ func (w *Worker) operate() error {
...
+			// sleep for the configured time
+			time.Sleep(w.Config.CheckIn)
+
+			// set checking time to now and call the server
+			registryWorker.SetLastCheckedIn(time.Now().UTC().Unix())
+			_, _, err := w.VelaClient.Worker.Update(registryWorker.GetHostname(), registryWorker)
+
+			// if unable to update the worker, log the error but allow the worker to continue running
+			if err != nil {
+				logrus.Error("unable to update worker in database: %w", err)
```
jbrockopp, line 54: I think we could provide more information in this error message 👍 

Also, we'll want to remove the `%w` as that is a formatting directive only supported in the `fmt` pkg.

To be consistent, it should probably look identical to the [other error message that was written](https://github.com/go-vela/worker/blob/9438a4ab47b5a43d1a08ec0295f40e39b8286ec9/cmd/vela-worker/register.go#L42).

```suggestion
				logrus.Errorf("unable to update worker %s on the server: %v", registryWorker.GetHostname(), err)
```

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,44 @@
...
+	"net/http"
+
+	"github.com/go-vela/types/library"
+	"github.com/sirupsen/logrus"
+)
+
+// register is a helper function to register
+// the worker in the database, updating the item
+// if the worker already exists
+func (w *Worker) register(config *library.Worker) {
```
jbrockopp, line 17: I think we should have this function return an `error` 👍 

```suggestion
func (w *Worker) register(config *library.Worker) error {
```

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,44 @@
...
+	// check to see if the worker already exists in the database
+	_, resp, err := w.VelaClient.Worker.Get(config.GetHostname())
+	if resp.StatusCode == http.StatusNotFound {
+		// worker does not exist, create it
+		logrus.Infof("registering worker %s with the server", config.GetHostname())
+		_, _, err := w.VelaClient.Worker.Add(config)
+		if err != nil {
+			// log the error instead of returning so the operation doesn't block worker deployment
+			logrus.Errorf("unable to register worker %s with the server: %v", config.GetHostname(), err)
```
jbrockopp, line 26: Return the error message here 👍 

```suggestion
			return fmt.Errorf("unable to register worker %s with the server: %v", config.GetHostname(), err)
```

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+		return
+	}
+
+	// if there was an error other than a 404, log the error.
+	if err != nil {
+		// log the error instead of returning so the operation doesn't block worker deployment
+		logrus.Errorf("unable to get worker %s from server: %v", config.GetHostname(), err)
```
jbrockopp, line 34: Return the error message here 👍 

```suggestion
		return fmt.Errorf("unable to get worker %s from server: %v", config.GetHostname(), err)
```

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	// the worker exists in the db, update it with the new config
+	logrus.Info("worker previously registered with server, updating information")
+	_, _, err = w.VelaClient.Worker.Update(config.GetHostname(), config)
+	if err != nil {
+		// log the error instead of returning so the operation doesn't block worker deployment
+		logrus.Errorf("unable to update worker %s on the server: %v", config.GetHostname(), err)
```
jbrockopp, line 42: Return the error message here 👍 

```suggestion
		return fmt.Errorf("unable to update worker %s on the server: %v", config.GetHostname(), err)
```

jbrockopp (changes requested):

cmd/vela-worker/operate.go
```diff
@@ -23,6 +26,35 @@ func (w *Worker) operate() error {
...
+			// sleep for the configured time
+			time.Sleep(w.Config.CheckIn)
+
+			// set checking time to now and call the server
+			registryWorker.SetLastCheckedIn(time.Now().UTC().Unix())
+			_, _, err := w.VelaClient.Worker.Update(registryWorker.GetHostname(), registryWorker)
+
+			// if unable to update the worker, log the error but allow the worker to continue running
+			if err != nil {
+				logrus.Errorf("unable to update worker %s in database: %v", registryWorker.GetHostname(), err)
```
jbrockopp, line 54: I believe the error message for this line and [this other line](https://github.com/go-vela/worker/blob/worker_table/cmd/vela-worker/register.go#L42) are still different.

Should they be?

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,46 @@
...
+		// log the error instead of returning so the operation doesn't block worker deployment
+		return fmt.Errorf("unable to get worker %s from server: %v", config.GetHostname(), err)
+	}
+
+	// the worker exists in the db, update it with the new config
+	logrus.Infof("worker %s previously registered with server, updating information", config.GetHostname())
+	_, _, err = w.VelaClient.Worker.Update(config.GetHostname(), config)
+	if err != nil {
+		// log the error instead of returning so the operation doesn't block worker deployment
+		return fmt.Errorf("unable to update worker %s on the server: %v", config.GetHostname(), err)
```
jbrockopp, line 42: I believe the error message for this line and [this other line](https://github.com/go-vela/worker/blob/worker_table/cmd/vela-worker/operate.go#L54) are still different.

Should they be?

jbrockopp (changes requested):

cmd/vela-worker/flags.go
```diff
@@ -26,6 +26,13 @@ func flags() []cli.Flag {
 			Usage:   "set hostname for the worker",
 		},
 
+		&cli.DurationFlag{
+			EnvVars: []string{"WORKER_CHECK_IN", "VELA_CHECK_IN", "CHECK_IN"},
+			Name:    "checkIn",
+			Usage:   "time to wait in between checking in with the server",
+			Value:   15,
```
wass3r, line 33: i think you'll want to use a unit here, something like `15 * time.Minute`

meinenec: Yes. Turns out it defaults to nanoseconds. Good catch!

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// register is a helper function to register
+// the worker in the database, updating the item
+// if the worker already exists
+func (w *Worker) register(config *library.Worker) error {
+	// check to see if the worker already exists in the database
+	_, resp, err := w.VelaClient.Worker.Get(config.GetHostname())
+	if err != nil {
+		// check to see if the response was nill
```
jbrockopp, line 22: I think this is a typo

```suggestion
		// check to see if the response was nil
```

cmd/vela-worker/register.go
```diff
@@ -0,0 +1,48 @@
...
+		if resp.StatusCode == http.StatusNotFound {
+			logrus.Infof("registering worker %s with the server", config.GetHostname())
+			_, _, err := w.VelaClient.Worker.Add(config)
+			if err != nil {
+				// log the error instead of returning so the operation doesn't block worker deployment
+				return fmt.Errorf("unable to register worker %s with the server: %v", config.GetHostname(), err)
+			}
+		} else {
+			return fmt.Errorf("unable to retrieve worker %s from the server: %v", config.GetHostname(), err)
+		}
```
jbrockopp, line 36: Do we want to `return nil` after we add the worker?

If I'm reading the code correctly, we attempt to get the worker to see if it exists.

If it doesn't exist, we check to see if the status code returned was a `404`(not found):
  * if status code == `404`, then we attempt to add the worker to Vela
  * if status code != `404`, then we return an error

I think this all sounds good, but the way it is written, I believe that after we successfully add the worker, then we'll exit out of the `if` block and continue processing the rest of this function.

The end result would be after we make the call to add the worker, we would then continue to the logic that updates the worker at the end of this function meaning that we make back to back API calls.

I know this is a book so I think I'll show what I'm suggesting:

```suggestion
		// check to see if the worker was not found and if we need to add it
		if resp.StatusCode == http.StatusNotFound {
			logrus.Infof("registering worker %s with the server", config.GetHostname())
			_, _, err := w.VelaClient.Worker.Add(config)
			if err != nil {
				// log the error instead of returning so the operation doesn't block worker deployment
				return fmt.Errorf("unable to register worker %s with the server: %v", config.GetHostname(), err)
			}
			
			// successfully added the worker so return nil
			return nil
		}
		
		return fmt.Errorf("unable to retrieve worker %s from the server: %v", config.GetHostname(), err)
```

jbrockopp (changes requested):

cmd/vela-worker/register.go
meinenec: Good catch! You looks to be correct 🤕

jbrockopp (approved):
LGTM

wass3r (approved):

PR #127, opened 2020-11-19
feat: add option to run with TLS

meinenec:
Allows serving over https if certs are provided.

### DIFF
--- a/cmd/vela-worker/flags.go
+++ b/cmd/vela-worker/flags.go
@@ -84,6 +84,16 @@ func flags() []cli.Flag {
 			Name:    "server.secret",
 			Usage:   "secret used for server <-> worker communication",
 		},
+		&cli.StringFlag{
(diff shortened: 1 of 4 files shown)


kneal (commented):
Does this limit someone from being able to run with an HTTP configuration?

Like if I'm running a local stack or can I run HTTP or can I chose to deploy it without HTTPS?

meinenec:
> Does this limit someone from being able to run with an HTTP configuration?
> 
> Like if I'm running a local stack or can I run HTTP or can I chose to deploy it without HTTPS?

Yup! That's the point of checking if certs are provided.
https://github.com/go-vela/worker/blob/3d0dc1335f93aa1001ecdbe7e4e14e500b0a978f/cmd/vela-worker/server.go#L43

kneal:
Awesome thanks for the clarification!

kneal (approved):
LGTM 🐬

davidvader (approved):

PR #130, opened 2020-12-28
fix: allow admin to define api protocol and set worker address

meinenec:
Fixes a bug where the worker's addr is being set to the server addr
https://github.com/go-vela/worker/blob/master/cmd/vela-worker/operate.go#L33

Changes the logic on running with TLS behind the new API_PROTOCOL flag instead of running with TLS if certs are present.


cmd/vela-worker/operate.go
```diff
@@ -26,11 +28,14 @@ func (w *Worker) operate() error {
 	if err != nil {
 		return err
 	}
+	// ensure the address has one and only instance of "://"
+	addr := fmt.Sprintf("%s://%s", strings.TrimSuffix(w.Config.API.Protocol, "://"), w.Config.Hostname)
```
matt-fevold, line 32: TIL [strings.TrimSuffix](https://golang.org/pkg/strings/#TrimSuffix) - for Java I would have done a .replace on that char sequence with an empty string, but this is a prettier way of that!

cmd/vela-worker/server.go
```diff
@@ -37,23 +38,25 @@ func (w *Worker) server() error {
...
-		_, err := os.Stat(w.Config.Certificate.Cert)
-		if err != nil {
-			logrus.Fatalf("Expecting certificate file at %s, got %v", w.Config.Certificate.Cert, err)
-		}
-		_, err = os.Stat(w.Config.Certificate.Key)
-		if err != nil {
-			logrus.Fatalf("Expecting certificate key at %s, got %v", w.Config.Certificate.Key, err)
+	// if running with HTTPS, check certs are provided and run with TLS.
+	if strings.EqualFold(w.Config.API.Protocol, "https") {
+		if len(w.Config.Certificate.Cert) < 0 && len(w.Config.Certificate.Key) < 0 {
```
matt-fevold, line 43: I think you have your `<` backwards here? Looks like it'll always equate to `if false && false` unless I'm reading it wrong :)

meinenec: Sure do. Forgot to change that back after testing something. Good catch!

meinenec:
Going to refactory slightly and try out the [url package](https://golang.org/pkg/net/url) to condense it down to just one variable for scheme, hostname, and port.

matt-fevold:
@meinenec should I hold off on my rereview until your refactor?

meinenec:
> @meinenec should I hold off on my rereview until your refactor?

yeah sorry, the holiday caught me off guard. The change is just different enough where I would recommend waiting 😃

matt-fevold:
All good @meinenec just wanted to check :)

meinenec:
Made a change to replace the api variables (scheme, hostname, and port) to all be included in the worker structs [config.API.Address](https://github.com/go-vela/worker/blob/e82f48547c84d80d08eae0ca47123ebc0378e6c6/cmd/vela-worker/worker.go#L20) variable.

We can access those fields with
- `.Hostname()` (`.Host` would give you `<hostname>:<port>` ie `localhost:8080`
- `.Port`
- `.Scheme`
and others

I thought it would be fair to keep this, https://github.com/go-vela/worker/blob/e82f48547c84d80d08eae0ca47123ebc0378e6c6/cmd/vela-worker/main.go#L20-L34

and use it similarly here https://github.com/go-vela/worker/blob/e82f48547c84d80d08eae0ca47123ebc0378e6c6/cmd/vela-worker/run.go#L135

meinenec:
For those curious, an unspecified port will always default to `8080` whether it's `http://localhost`, `http://localhost:`, or `https://localhost`

cmd/vela-worker/server.go
```diff
@@ -35,25 +37,31 @@ func (w *Worker) server() error {
...
+			_, err := os.Stat(w.Config.Certificate.Cert)
+			if err != nil {
+				logrus.Fatalf("Expecting certificate file at %s, got %v", w.Config.Certificate.Cert, err)
+			}
+			_, err = os.Stat(w.Config.Certificate.Key)
+			if err != nil {
+				logrus.Fatalf("Expecting certificate key at %s, got %v", w.Config.Certificate.Key, err)
+			}
+		} else {
+			logrus.Fatal("Unable to run with TLS: No certificate provided")
```
jbrockopp, line 55: Thoughts on potentially un-capitalizing the error messages here?

I don't think we have a clear defined format but in general, I believe we try using lowercase whenever possible 👍 

As an example:

```
# I think we want to do thiis
fmt.Errorf("sample lowercase error message")
logrus.Fatal("sample lowercase log message")

# I don't think we want to do this
fmt.Errorf("Sample capitalized error message")
logrus.Fatal("Sample capitalized log message")
```

cmd/vela-worker/validate.go
```diff
@@ -17,6 +19,19 @@ func (w *Worker) Validate() error {
...
+	// check that hostname was properly populated
+	if len(w.Config.API.Address.Hostname()) == 0 {
+		switch strings.ToLower(w.Config.API.Address.Scheme) {
+		case "http", "https":
+			retErr := "Worker server address invalid: %s"
+			return fmt.Errorf(retErr, w.Config.API.Address.String())
+		default:
+			// hostname will be empty if a scheme is not provided
+			retErr := "Worker server address invalid, no scheme: %s"
+			return fmt.Errorf(retErr, w.Config.API.Address.String())
```
jbrockopp, line 31: Thoughts on potentially un-capitalizing the error messages here?

jbrockopp (changes requested):

jbrockopp (dismissed):
LGTM

cmd/vela-worker/exec.go
```diff
@@ -19,6 +19,7 @@ import (
 
 // exec is a helper function to poll the queue
 // and execute Vela pipelines for the Worker.
+// nolint:funlen
```
jbrockopp, line 22: Could we add a description for the `nolint` comment here?

Perhaps something like:

```suggestion
// nolint:funlen // ignore function length due to comments
```

jbrockopp (changes requested):

cmd/vela-worker/exec.go
meinenec: We can add something, but I don't think it's the comments that are making the function too long since the function is over 100 lines with comments and the threshold is at 50. Do you have any suggestions?

jbrockopp: FYI:

I think the threshold for lines is actually set at `100` based off our configuration:

https://github.com/go-vela/worker/blob/45391b6040590d19c779da539e5f0e7acd8fab73/.golangci.yml#L29-L32

The linter source code can be found here:

https://github.com/ultraware/funlen

It measures both lines and statements and I think the issue here was about the statements [based off the Actions output](https://github.com/go-vela/worker/runs/1663678947):

image

All that being said, I can see what you're talking about maybe it's not only comments.

Perhaps we also blame log messages as well? 😅 

```suggestion
// nolint:funlen // ignore function length due to comments and log messages
```

meinenec: I think that's a fair comment 😄 

Thank you for digging into that and providing more info than I did. I'm going to steal that comment for now!

wass3r (approved):

jbrockopp (approved):
LGTM

PR #137, opened 2021-01-11
fix: remove nil check from worker registration

meinenec:
Removing a check that isn't currently checking anything. If we remove this check the error is still caught below. Currently this is causing an issue when attempting to add a worker that doesn't yet exist in the db.


jbrockopp (approved):
LGtM

KellyMerrick (approved):
lgtm

PR #142, opened 2021-01-21
fix: allow workers to retry registering with the server

meinenec:
mades the registration process a little more readable and allows the worker to retry registering if the worker was not previously registered. This removes a race condition where the server must be available before the worker

fixes https://github.com/go-vela/community/issues/185


matt-fevold (approved):

jbrockopp (approved):
LGTM

PR #152, opened 2021-02-04
fix(cancel): return build struct instead of string resp

meinenec:
Returns the build struct instead of a string saying "canceled build X" so the api has more context of the build canceled.


davidvader (approved):

jbrockopp (approved):
LGTM

PR #169, opened 2021-02-22
fix(executors): set id in returned payload

JordanSussman:
This change is needed to allow the server to properly determine which executor contains the relevant build to cancel. Relevant code on the server includes:

- `executor.GetID()` from https://github.com/go-vela/server/blob/5667aeb069d371df1aa7bfbe72a28acbeebcb5b8/api/build.go#L1149
- Adding executors to context https://github.com/go-vela/server/blob/5667aeb069d371df1aa7bfbe72a28acbeebcb5b8/router/middleware/executors/executors.go#L87

closes https://github.com/go-vela/community/issues/217


KellyMerrick (approved):
lgtm

PR #232, opened 2021-11-03
revert: stream container logs

jbrockopp:
Based off of https://github.com/go-vela/community/blob/master/proposals/2021/07-22_log-streaming.md

Related to https://github.com/go-vela/community/issues/367 and https://github.com/go-vela/community/issues/368

Essentially, this is a revert of https://github.com/go-vela/pkg-executor/pull/177

This reverts the changes previously introduced in the `v0.10.x` release to stream container logs.

An issue was discovered with the current implementation of streaming logs:

https://github.com/go-vela/community/issues/427

In a later PR, we'll look to introduce a [different implementation of streaming logs](https://github.com/go-vela/community/blob/master/proposals/2021/07-22_log-streaming.md#option-1):

* https://github.com/go-vela/pkg-executor/tree/feature/log_streaming/opt_one

However, this functionality will be behind a feature flag and be considered opt-in.

### DIFF
--- a/executor/linux/service.go
+++ b/executor/linux/service.go
@@ -5,6 +5,8 @@
 package linux
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
@@ -175,6 +177,9 @@ func (c *client) StreamService(ctx context.Context, ctn *pipeline.Container) err
 		return err
 	}
 
+	// create new buffer for uploading logs
+	logs := new(bytes.Buffer)
+
 	// nolint: dupl // ignore similar code
 	defer func() {
 		// tail the runtime container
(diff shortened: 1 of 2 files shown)


cognifloyd:
Can't we adjust the load balancer settings to allow streaming instead of removing this?  Can we look at how kubernetes handles log streaming, because that works quite well and is also written in go. For my Vela instance in kubernetes, I often use `stern` to stream the logs from pipeline comedones faster than they show up in the Vela UI.

jbrockopp:
@cognifloyd I want to make sure we're on the same page before answering your question 👍 

The plan of attack is to revert the existing implementation of streaming logs from the server -> worker temporarily.

I'm currently working on finalizing code locally to implement the streaming logic from option 1 in this PR description.

This will still allow streaming logs but won't be susceptible to issues depending on architecture setup with Vela.

i.e. load balancer configuration

jbrockopp:
> Can't we adjust the load balancer settings to allow streaming instead of removing this? Can we look at how kubernetes handles log streaming, because that works quite well and is also written in go. For my Vela instance in kubernetes, I often use `stern` to stream the logs from pipeline comedones faster than they show up in the Vela UI.

We could certainly attempt to remediate the problem by updating the load balancer.

However, determining what settings to apply to the load balancer may get tricky.

For example, AWS uses a default idle connection timeout of `60s` which I'm told is fairly common.

What would make sense to set as the new timeout? `5m`? `10m`?

And even if we did apply that configuration, there would be still be the potential for a consumer to face issues.

To clarify, I could set the idle connection timeout at `5m`.

But, any pipeline that didn't send logs in `5m` or less time would face the same issue.

The easiest way I could think to reproduce that issue would be:

```yaml
version: "1"

steps:
  - name: sleep
    image: alpine:latest
    commands:
      - sleep 301
      - echo foo
```

Another point to mention is the possibility for admins of Vela to not have full control of the load balancer configuration.

This would mean they would be unable to make the changes necessary for their load balancer.

jbrockopp:
Closing this in favor of https://github.com/go-vela/worker/pull/232

PR #235, opened 2021-11-05
fix: compare time chunks against zero

davidvader:
fixes a bug in the linux executor where timed chunks are not written until the buffer becomes larger than the size of the log itself.


wass3r (approved):

jbrockopp (approved):
LGTM

PR #239, opened 2021-11-13
fix: move Step init logic to library

cognifloyd:
Both the local and linux executors have to create a step object in a couple of places. This moves that logic into `library` (types) to facilitate reuse.

The executor can't use `StepFromContainer` because that relies on the container Environment which is not configured yet.  So, we rename `StepFromContainer` to `StepFromContainerEnvironment` to more accurately represent what it does. Then we add a new `StepFromBuildContainer` function to create a new Step and initialize it using data from a `library.Build` and a `pipeline.Container`.

For background, see the conversation here: https://github.com/go-vela/pkg-executor/pull/179#discussion_r735875710

This depends on https://github.com/go-vela/types/pull/205

- StepFromContainer -> StepFromContainerEnvironment
- Use library.StepFromBuildContainer to create new Steps


cognifloyd:
marking this as draft until https://github.com/go-vela/types/pull/205 is merged

wass3r:
i think we're good to open this up now

cognifloyd:
I'm not near a computer right now. The only thing left for this PR is to update the types module

jbrockopp (changes requested):
Think we'll need to vendor in the latest changes from [go-vela/types](https://github.com/go-vela/types) for this to work properly:

`go get -u github.com/go-vela/types@master`

cognifloyd:
K. This is ready :)

wass3r (approved):

jbrockopp (approved):
LGTM

PR #246, opened 2021-12-03
enhance: use application logger

jbrockopp:
part of the effort for https://github.com/go-vela/community/issues/419

This change updates the worker to have a consistent, standardized set of fields when running a build.

To achieve this, we create a logger at the application level with the standard set of fields:

https://github.com/go-vela/worker/blob/57f21ed56d2e24a949d2e84babd469214e78f9b9/cmd/vela-worker/exec.go#L38-L49

And pass that logger to the `executor` and `runtime` packages:

https://github.com/go-vela/worker/blob/57f21ed56d2e24a949d2e84babd469214e78f9b9/cmd/vela-worker/exec.go#L51-L55

https://github.com/go-vela/worker/blob/57f21ed56d2e24a949d2e84babd469214e78f9b9/cmd/vela-worker/exec.go#L66-L70

## Log Samples

Here is a snippet of what the logs look like before this change:

```
{"build":1,"host":"worker","level":"info","msg":"creating build","repo":"JordanBrockopp/vela-play","time":"2021-12-13T20:32:22Z","version":"v0.11.0"}
{"build":1,"host":"worker","level":"info","msg":"uploading build state","repo":"JordanBrockopp/vela-play","time":"2021-12-13T20:32:22Z","user":"JordanBrockopp"}
{"level":"trace","msg":"no-op: setting up for build JordanBrockopp_vela-play_1","time":"2021-12-13T20:32:22Z"}
{"build":1,"host":"worker","level":"info","msg":"creating init step","repo":"JordanBrockopp/vela-play","time":"2021-12-13T20:32:22Z","user":"JordanBrockopp"}
{"build":1,"host":"worker","level":"info","msg":"planning init step","repo":"JordanBrockopp/vela-play","time":"2021-12-13T20:32:22Z","user":"JordanBrockopp"}
{"build":1,"host":"worker","level":"debug","msg":"uploading step state","repo":"JordanBrockopp/vela-play","step":"init","time":"2021-12-13T20:32:22Z","user":"JordanBrockopp"}
{"build":1,"host":"worker","level":"debug","msg":"retrieve step log","repo":"JordanBrockopp/vela-play","step":"init","time":"2021-12-13T20:32:22Z","user":"JordanBrockopp"}
{"build":1,"host":"worker","level":"debug","msg":"uploading build snapshot","repo":"JordanBrockopp/vela-play","time":"2021-12-13T20:32:22Z","user":"JordanBrockopp"}
[...]

PR #253, opened 2021-12-15
fix(docker): pulling images without output

jbrockopp:
Based off of https://github.com/go-vela/worker/pull/246

An issue was discovered with the `target/vela-worker:latest` Docker image produced from this codebase.

We pulled that version into our lower environments today for testing and noticed a discrepancy:

image

The issue seems specifically isolated to when the worker attempts to pull Docker images.

SEE EDIT - ~Thus far, was unsuccessful in reproducing locally but it could be related to us using an internal registry mirror.~ 

At any rate, was able to reproduce the issue on our internal infrastructure and this seems to resolve that issue.

## EDIT

Thanks to [a comment made](https://github.com/go-vela/worker/pull/253#issuecomment-995163389) by @wass3r, I was able to figure out how to reproduce it locally.

[In the above PR](https://github.com/go-vela/worker/pull/246), a change was made to ignore the output from the `docker pull` performed by a worker.

The exception to this, where you will see this output, is when the worker is setup with `trace` level logging:

https://github.com/go-vela/worker/blob/c123efa33438da8063df81189d0bb43fbe0a5504/runtime/docker/image.go#L47-L54

This change inadvertently introduced a bug when the worker attempts to fetch a new Docker image for a build.

You can reproduce this issue with the following steps:

1. Checkout to the `master` branch of this repository:

```shell
$ git checkout master
```

2. Remove `trace` level logging for the worker by removing or commenting out the following line:

https://github.com/go-vela/worker/blob/c123efa33438da8063df81189d0bb43fbe0a5504/docker-compose.yml#L29

3. Remove a Docker image from your local image cache so the worker is forced to pull it when a build runs:

```shell
$ docker rmi -f target/vela-git:v0.4.0
```

4. Start the [local Vela stack](https://github.com/go-vela/worker/blob/master/DOCS.md) via Docker-Compose.

5. Run a build locally and watch the error message above be displayed in the Vela UI.

### DIFF
--- a/runtime/docker/image.go
+++ b/runtime/docker/image.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"strings"
 
@@ -48,6 +49,12 @@ func (c *client) CreateImage(ctx context.Context, ctn *pipeline.Container) error
 	if logrus.GetLevel() == logrus.TraceLevel {
 		// copy output from image pull to standard output
 		_, err = io.Copy(os.Stdout, reader)
+		if err != nil {
+			return err
+		}
+	} else {
+		// discard output from image pull
+		_, err = io.Copy(ioutil.Discard, reader)
 		if err != nil {
 			return err
 		}


wass3r:
if i understand this change right, you'd would still see this with trace logging. is that correct?

jbrockopp:
@wass3r no its actually the inverse 😅 

Your comment inspired a thought bubble that allowed me to come up with a way to reproduce it locally.

You don't see this bug when you use `trace` level logging which is why I couldn't reproduce it locally:

https://github.com/go-vela/worker/blob/c123efa33438da8063df81189d0bb43fbe0a5504/docker-compose.yml#L29

Our internal infrastructure that is setup doesn't use that level of logging for the worker.

I'll update my original description accordingly with further details.

jbrockopp:
@wass3r the PR description has been updated 😃

wass3r (approved):

JordanSussman (approved):

PR #254, opened 2021-12-21
feat(secrets): mask secrets in logs

ecrupper:
[Related Issue](https://github.com/go-vela/community/issues/136)

- Added a replace routine to the `step.go` to intercept secret-laden logs before they get updated in the database and mask the secrets.

## Current Situation
Screen Shot 2021-12-21 at 2 45 02 PM

## With Change
Screen Shot 2021-12-21 at 2 47 08 PM


wass3r:
is there an edge case here where it's possible that the buffer contains the partial secret (whether time chunks or byte chunks) and won't properly get masked - i know the final log upload from the defer will take care of it, but seems like there is a limitation here? another scenario i am wondering about is multi-line secrets - did you do some testing there as well?

executor/linux/step.go
```diff
@@ -214,6 +215,13 @@ func (c *client) StreamStep(ctx context.Context, ctn *pipeline.Container) error
 		return err
 	}
 
+	secretValues := []string{}
+
+	for _, secret := range ctn.Secrets {
+		s := ctn.Environment[strings.ToUpper(secret.Target)]
+		secretValues = append(secretValues, s)
+	}
```
jbrockopp, line 223: Does this implementation account for external `secrets`?

https://go-vela.github.io/docs/tour/secrets/

https://go-vela.github.io/docs/usage/examples/secrets_external/

https://go-vela.github.io/docs/reference/yaml/secrets/#the-origin-tag

These secrets are unique since they are fetched by a Docker container and written to the build workspace.

AFAICT, it appears this checks for internal `secrets` declared in the pipeline for the `step`:

https://go-vela.github.io/docs/usage/examples/secrets_internal/

If possible, I think we should try to account for external secrets declared in the pipeline as well.

jbrockopp (changes requested):

executor/linux/step.go
wass3r: that's a good point, too. they should all be in `/vela/secrets/` (at least for how it currently works).

ecrupper:
Decided to pull out the gathering of secret values and the masking of logs into their own helper functions for easier testing. In the process of doing this, I discovered that secrets which contain a substring of another secret will not be handled properly, so I introduced some RegExp logic to address that. Will work on incorporating external secrets now.

executor/linux/step.go
kneal: I'm not sure we will be able to catch external secrets because the vela worker code won't be running in the workspace that the `/vela/secrets` would exist within. 

It's possible you could make calls to the Docker volume to get them but I'm not sure we have those functions within the runtime.

kneal: To elaborate (kinda just word vomited), Vela external secrets are entirely controller by plugins meaning the worker's responsibility, in that case, is just managing the container, and the secrets are written to the dynamic volume that is created. 

So, if the worker wanted to see what those values are it would need to interrogate the volume that was spun up for that particular build and then get the values and do its sanitization. I'm not saying it's impossible but I don't think we are set up runtime-wise yet to make calls to the volumes to analyze content within them.

ecrupper: With that being said, should this PR be only internal secrets? Create a separate issue for external secret masking?

kneal: I would vote scope to internal but we should hear from others. I think external secrets need to be re-evaluated again. 😢 

We know a lot more now than when we did the first couple of attempts at writing the feature and I think it needs to be adapted.  I've seen many requests to make them ENV while I'm not sure that's entirely possible we could look into it again. That would potentially give us more options for sanitization but I don't know at this point.

jbrockopp: Yeah, I wasn't sure if it was possible or not 🤷 

Was thinking we could try and add code that inspects the build workspace (`/vela`) to get the secret values?

If everyone is on board for keeping this feature scoped to `internal` secrets I think I can get behind that 😃 

We can always worry about this functionality later for `external` secrets

Could I request a story is written for it though?

i.e. mask `external` secret in the logs

ecrupper: [External Secrets Masking Issue](https://github.com/go-vela/community/issues/465)

executor/linux/step_test.go
```diff
@@ -338,9 +340,6 @@ func TestLinux_StreamStep(t *testing.T) {
 	}
 
 	_runtime, err := docker.NewMock()
-	if err != nil {
-		t.Errorf("unable to create runtime engine: %v", err)
-	}
```
jbrockopp, line 343: Could you share why these lines were removed?

Perhaps it was an accident? 😅

jbrockopp (changes requested):

executor/linux/step_test.go
ecrupper: Yes it was 😅

executor/linux/step.go
```diff
@@ -438,3 +451,39 @@ func (c *client) DestroyStep(ctx context.Context, ctn *pipeline.Container) error
...
+		re := regexp.MustCompile(`\s` + secret + `\s`)
+		matches := re.FindAllString(strData, -1)
+		for _, match := range matches {
+			mask := string(match[0]) + constants.SecretLogMask + string(match[len(match)-1])
+			strData = strings.Replace(strData, match, mask, 1)
+		}
+		strData = re.ReplaceAllString(strData, constants.SecretLogMask)
+	}
+	return []byte(strData)
+}
```
jbrockopp, line 489: Do we think this should be a method on the `github.com/go-vela/library/Log{}` type?

Thinking the function signature could look like:

```go
func (l *Log) MaskData(secrets []string) {}
```

Then, to use this function in the the code above:

```go
// update the existing log with the new bytes
//
// https://pkg.go.dev/github.com/go-vela/types/library?tab=doc#Log.AppendData
_log.AppendData(logs.Bytes())

// mask any secrets in the log data
//
// https://pkg.go.dev/github.com/go-vela/types/library?tab=doc#Log.MaskData
_log.MaskData(secretValues)
```

ecrupper: That might be helpful in the eventual effort to have this masking apply to external secrets

jbrockopp: You read my mind 👍  I was thinking along those same lines!

Also, if/when we write out other executors then we don't have to duplicate code.

In the interest of not blocking the functionality, I'll approve for now 😄

jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

PR #258, opened 2022-01-19
enhance(spec): automate worker api spec generation

wass3r:
this change follows in the footsteps of go-vela/server with regards to api spec generation. going forward, the spec will be generated, validated, and uploaded as an artifact as part of any release.


jbrockopp (approved):
LGTM

KellyMerrick (approved):
lgtm

PR #261, opened 2022-01-19
chore(step): clean up masking routine due to types change

ecrupper:
Since [masking is now done as a log method](https://github.com/go-vela/types/pull/225), we don't need to have it done within `StreamStep`.

Also edited the method names from `ServiceFromContainer` to `ServiceFromContainerEnvironment` due to some breaking types changes I pulled in.


jbrockopp (approved):
LGTM

wass3r (approved):

PR #271, opened 2022-02-11
fix(logs): creating a flag for secret masking

ecrupper:
Some weird behavior has occurred with the secret masking functionality released with `v0.12.0`. There are some end users presumably with secret values that were not anticipated, so while we address that, I think buying us some time with this feature flag is a good idea.


jbrockopp (changes requested):
I think we can close this in favor of:

https://github.com/go-vela/worker/pull/272

PR #272, opened 2022-02-14
fix(step): add catch block for disallowed secrets

ecrupper:
When pulling secret values to be masked, we need to catch the instance where a secret is not allowed for that event. Otherwise, when the map does not find the secret, it returns an empty string.


jbrockopp (changes requested):
Per our discussion in Slack, this code change LGTM 👍 

Could we maybe add a test or two to help account for this behavior?

jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

PR #274, opened 2022-02-18
fix(logs/secrets): check for disallowed/non-existent secrets

ecrupper:
Backport for patch release (v0.12.1). 

When users inject secrets for a container that are not allowed for the event or simply do not exist, the `getSecretValues` function would set them to an empty string, resulting in ugly logs. 

Earlier, merged PR: https://github.com/go-vela/worker/pull/272


jbrockopp (approved):
LGTM

wass3r (approved):

PR #277, opened 2022-02-24
fix(kubernetes): Make Executor TailContainer before RunContainer

cognifloyd:
This should make Kubernetes log capture more robust without changing the behavior for the Docker runtime.

The Kubernetes runtime needs to start asking for logs before the container is created, or logs from short-lived steps might be gone by the time TailContainer starts.

The Docker runtime, however, needs to RunContainer first or the docker engine will complain that the container does not exist. So, we use a special channel to let the Docker Runtime's TailContainer know when RunContainer has finished to allow it to delay pulling logs until the container exists.


cognifloyd:
This makes use of a new `CancelContext` to inform `TailContainer` once `RunContainer` has finished.

Is there a better signaling primitive to synchronize the stream+TailContainer go routine with when the outer scope finishes RunContainer?

cognifloyd:
Oops. I hit the wrong button and closed it. :(

cognifloyd:
I switched from a Context to a plain channel. That feels a bit cleaner.

executor/linux/stage.go
```diff
@@ -31,7 +31,7 @@ func (c *client) CreateStage(ctx context.Context, s *pipeline.Stage) error {
 	// update the init log with progress
 	//
 	// https://pkg.go.dev/github.com/go-vela/types/library?tab=doc#Log.AppendData
-	_log.AppendData([]byte(fmt.Sprintf("> Pulling step images for stage %s...\n", s.Name)))
+	_log.AppendData([]byte(fmt.Sprintf("> Pulling step images for stage %s...\n", s.Name))) // TODO: "Pulling" only makes sense for Docker runtime
```
cognifloyd, line 34:
```suggestion
	_log.AppendData([]byte(fmt.Sprintf("> Pulling step images for stage %s...\n", s.Name)))
```

cognifloyd:
Alternate version in #286

cognifloyd:
Closing in favor of #303

PR #279, opened 2022-02-26
feat(kubernetes): Surface Image Pull Errors in Kubernetes Runtime

cognifloyd:
The Docker Runtime surfaces image pull errors naturally because of the blocking synchronous requests to pull an image or start a container.
Kubernetes is asynchronous, so we have to explicitly watch for events to make sure the image pull succeeds in RunContainer.
When we capture a failure event, we can surface the event in the same way as the Docker runtime.

There may be other places where we need to watch for events, so I extracted some of the test logic from WaitContainer to make it reusable.

Also, this changes the init step so it says "Preparing" instead of "Pulling" since the Kubernetes runtime does NOT pull during init.

- Replace "Pulling" with "Preparing" in init step logs
- More image validation in kubernetes runtime
- Rename watch->podWatch for planned imports
- Extract Kubernetes Client + Watch mocking logic
- Escape selector just in case
- Move newMockWithWatch to kubernetes_test.go
- Close the podWatch in WaitContainer
- Enhance RunContainer to watch for image pull events
- Add RunContainer image pull failure test


runtime/kubernetes/image.go
```diff
@@ -29,8 +30,21 @@ const imagePatch = `
...
+	//
+	// https://pkg.go.dev/github.com/go-vela/worker/internal/image#ParseWithError
+	_, err := image.ParseWithError(ctn.Image)
+	if err != nil {
+		return err
+	}
+
+	// Kubernetes does not have an API to make sure it can access an image,
+	// so we have to query the appropriate docker registry ourselves.
+	// TODO: query docker registry for the image (if possible)
```
kneal, line 45: Does it make sense to pull this TODO out into a story?

kneal (dismissed):
LGTM 🐬

runtime/kubernetes/image.go
cognifloyd: Yeah. That'll be involved because we need whatever credentials are used with the (private in my case) docker registry. So either pull it from k8s or the admin has to configure Vela with it.

Do you mean put it in an issue on the community repo?

wass3r (approved):
thank you!

runtime/kubernetes/image.go
cognifloyd: By story, do you mean an issue on the community repo?

kneal: Yeah, sorry poor phrasing

wass3r:
can you please fix up the new linting errors? we added/modified some rules that triggered a few issues.

cognifloyd:
I'm really not sure what to do to replace `context.Background()` for the k8s no API requests. The other instances were skipped, stop that's what I did. Is there a better way?

cognifloyd:
I'm going to break this down into more discrete PRs. As I'm studying issues with the k8s runtime, I'm considering some different ways to subscribe to the event stream related to a build.

runtime/kubernetes/image.go
cognifloyd: Done: https://github.com/go-vela/community/issues/579

cognifloyd:
Closing in favor of #331

PR #280, opened 2022-03-01
enhance(http): add retry client

JordanSussman:
This aims to hopefully help workaround intermittent errors returned by the Vela by adding a retryable http client created by [go-retryablehttp](https://github.com/hashicorp/go-retryablehttp). Do folks think we should override the [default retry check](https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L386) to allow retrying on `context.DeadlineExceeded` errors? I ask this since prior connectivity failures reported like https://github.com/go-vela/community/issues/434 specifically had deadline exceeded errors.

### DIFF
--- a/cmd/vela-worker/client.go
+++ b/cmd/vela-worker/client.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"github.com/go-vela/sdk-go/vela"
+	"github.com/hashicorp/go-retryablehttp"
 
 	"github.com/sirupsen/logrus"
 )
@@ -13,11 +14,15 @@ import (
 // helper function to setup the queue from the CLI arguments.
 func setupClient(s *Server) (*vela.Client, error) {
 	logrus.Debug("creating vela client from worker configuration")
+	// create client to retryable http calls
+	retryClient := retryablehttp.NewClient()
+	// set logger to nil to avoid spam
+	retryClient.Logger = nil
 
 	// create a new Vela client from the server configuration
 	//
 	// https://pkg.go.dev/github.com/go-vela/sdk-go/vela?tab=doc#NewClient
-	vela, err := vela.NewClient(s.Address, "", nil)
+	vela, err := vela.NewClient(s.Address, "", retryClient.StandardClient())
 	if err != nil {
 		return nil, err
 	}


cmd/vela-worker/client.go
```diff
@@ -6,18 +6,23 @@ package main
...
 	"github.com/sirupsen/logrus"
 )
 
 // helper function to setup the queue from the CLI arguments.
 func setupClient(s *Server) (*vela.Client, error) {
 	logrus.Debug("creating vela client from worker configuration")
+	// create client to retryable http calls
+	retryClient := retryablehttp.NewClient()
+	// set logger to nil to avoid spam
+	retryClient.Logger = nil
```
JordanSussman, line 20: This seemed like the safest approach since the default behavior is to log **all** http requests regardless of success or failure.

kneal: I think I'm okay with it for now to see how it behaves, took me awhile to find [this](https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L47-L50) so I'm adding it here:
```go
// Default retry configuration
defaultRetryWaitMin = 1 * time.Second
defaultRetryWaitMax = 30 * time.Second
defaultRetryMax     = 4
```

runtime/docker/docker.go
```diff
@@ -71,10 +72,17 @@ func New(opts ...ClientOpt) (*client, error) {
...
+	// https://pkg.go.dev/github.com/hashicorp/go-retryablehttp#NewClient
+	retryClient := retryablehttp.NewClient()
+	// set logger to nil to avoid spam
+	retryClient.Logger = nil
+
 	// create new Docker client from environment
 	//
 	// https://godoc.org/github.com/docker/docker/client#NewClientWithOpts
-	_docker, err := docker.NewClientWithOpts(docker.FromEnv)
+	_docker, err := docker.NewClientWithOpts(docker.FromEnv, docker.WithHTTPClient(retryClient.StandardClient()))
```
kneal, line 85: Per the comment above, the `defaultRetryWaitMax` does feel kinda long IMO.

I'm just thinking about like what if the `ContainerCreate` call succeeds but I wait 30 seconds to bomb out for the `ContainerStart` I feel like in those sequential situations it's likely going to bomb out anyway so waiting 30 seconds feels a bit overboard. Thoughts?

JordanSussman: I'm also thinking about reverting this change since it likely has unintended consequences like what you mentioned. My initial thought for this was specifically making [ImagePull](https://github.com/go-vela/worker/blob/19081a02a335086f774e8a73f595e9efb873659d/runtime/docker/image.go#L42) fail less often when connecting to somewhat unreliable repositories - suppose I'll consider adding some specific retry logic to that particular function in another pull request.

kneal: Yeah, honestly, I think this package would have a ton of value in the sdk, but even then I would want to tune things because I don't think the `defaultRetryWaitMax` should be longer than like 10 seconds if its been added as a default to the entire client.

wass3r (commented):
I'm a bit concerned by this change, especially with regards to log pushes. since we do that unordered and async it will become more likely that logs will not be complete, correct?

contrived example with time stamps:
00:01 log chunk #1 push starts
00:02 log chunk #2 push starts
00:03 log chunk #1 fails, retrying
00:04 log chunk #2 push succeeds
00:05 log chunk #1 push succeeds (we will end up with a version of the logs in this state)

i know we do a final sweep to push logs (in the defer), but there again, it would be possible for the last or second to last log update to go through some retries while the final sweep already completed. i have no proof, but i think we're seeing some of that (rarely) today, but we might make that worse with this?

JordanSussman:
> I'm a bit concerned by this change, especially with regards to log pushes. since we do that unordered and async it will become more likely that logs will not be complete, correct?

I think you bring up a really valid argument against the retry logic at the global level since specific endpoints like log pushes could have unexpected results. Wondering if a decent middle ground would be utilizing the standard http library for logic specific endpoints and retry http library for everything else.

> i know we do a final sweep to push logs (in the defer), but there again, it would be possible for the last or second to last log update to go through some retries while the final sweep already completed. i have no solid proof, but i think we're seeing some of that (rarely) today, but we might make that worse with this?

I'm thinking the current behavior is more often caused by processing times when a large number of logs are written near the end of the build. For example, purpose the following behavior in a build:

1. Build starts
2. 100 lines are written in 100 seconds
3. 10,0000 lines are written in 1 second
4. 20 lines are written in 1 second
5. build ends

FWIW, the only builds we have seen the truncated log lines follow the behavior I described.

wass3r:
> I think you bring up a really valid argument against the retry logic at the global level since specific endpoints like log pushes could have unexpected results. Wondering if a decent middle ground would be utilizing the standard http library for log specific endpoints and retry http library for everything else.

I can get with that. I know the change won't be as clean as it is currently, but I would feel better about those adverse side effects.

> FWIW, the only builds we have seen the truncated log lines follow the behavior I described.

Yea, that's likely true. Notwithstanding network blips and such that is probably the most common scenario.

wass3rw3rk:
Getting back to this PR. What are the thoughts with byte-chunks being out of the picture now? I think the issue should be less pronounced but likely not eliminated, so a separation as @JordanSussman pointed out probably still makes sense?

JordanSussman:
@wass3r I'm fine with punting this change to specific endpoints in the future. FWIW we have not seen any major issues without this change since the byte-chunks was changed.

plyr4:
@JordanSussman hey we're going to close this PR as stale, but we definitely want better timeout resiliency so i'm going to create a separate issue for it and link to this PR as a great starting place.
https://github.com/go-vela/community/issues/863

PR #286, opened 2022-03-12
fix(kubernetes): Coordinate Runtime's RunContainer/TailContainer via a channel

cognifloyd:
The Kubernetes runtime needs to start streaming logs before the container starts.
The Docker runtime needs to start the container before streaming longs.
With both running in goroutines, the Runtime can block one method until the other completes in whichever order makes sense for that Runtime.

Alternate to #277


cognifloyd:
Closing in favor of #303

PR #294, opened 2022-03-18
feat(kubernetes): Add PipelinePodsTemplate CRD to define worker-specific Pod defaults for Kubernetes Runtime

cognifloyd:
Part of: https://github.com/go-vela/community/issues/515
In particular: https://github.com/go-vela/community/issues/515#issuecomment-1057652724

Documentation PR: https://github.com/go-vela/docs/pull/294

For every build:
1. load the PipelinePodsTemplate CRD from Kubernetes or from the filesystem (no need to restart the worker after update).
    - if the Vela admin has configured `VELA_RUNTIME_PODS_TEMPLATE_NAME` on the worker with the name of which PipelinePodsTemplate to use (from the `VELA_RUNTIME_NAMESPACE`),
        - The CRD gets in `runtime.SetupBuild()`
    - if the Vela admin **has not** configured `VELA_RUNTIME_PODS_TEMPLATE_NAME` but **has** configured `VELA_RUNTIME_TEMPLATE_FILE` on the worker with a path to a PipelinePodsTemplate yaml file on the local filesystem, then fall back to reading that file (ie: prefer reading CRD from kubernetes)
        - The CRD file gets read during runtime initialization
        - This allows a Vela admin to avoid loading the CRD if they do not want to or are not allowed to use CRDs in their cluster.
2. Then, use that object to populate various defaults when creating the build's Pipeline Pod
    - including labels, annotations, and SecurityContext, ...

This can be expanded to include additional Pod defaults in the future.

Each commit can be reviewed individually. I recommend reviewing them in order:

- feat(k8s): Define PipelinePodsTemplate CRD
- chore: Add crd generation targets to Makefile
- chore: Generate client for PipelinePodsTemplate CRD
- chore: Generate PipelinePodsTemplate CRD manifest
- feat(k8s): Add PipelinePodsTemplate loading opts
- feat(k8s): Get PipelinePodsTemplate from k8s API
- feat(k8s): Get defaults from PipelinePodTemplate
- the remaining commits focus on adding tests


cognifloyd:
I split and reorganized these commits out of a much larger experimental branch where I figured out how to create and use CRDs.

This still needs tests, but I'd like some feedback on the merit of this approach before I figure that out.

Makefile
```diff
@@ -311,3 +323,43 @@ spec-version-update:
...
+	cp -r ${TMP}/github.com/go-vela/worker/runtime/kubernetes/* runtime/kubernetes/
+	rm -rf $(TMP)
+	@echo "### crd clientset created successfully"
+
+# The `crd` target will call crd-gen to create a k8s crd
+# for the kubernetes runtime.
+#
+# Usage: `make crd`
+.PHONY: crd
+crd: crd-client-gen
```
cognifloyd, line 365: I need to add something that generates the CRD yaml file as well.

runtime/kubernetes/codegen/doc.go
```diff
@@ -0,0 +1,13 @@
...
+// Use of this source code is governed by the LICENSE file in this repository.
+
+// Package codegen is a dummy package to help go mod know that
+// we need k8s.io/code-generator.
+//
+// This package is not meant to be imported anywhere.
+package codegen
+
+import (
+	_ "k8s.io/code-generator"
```
cognifloyd, line 12: golangci is complaining:
> blank-imports: a blank import should be only in a main or test package, or have a comment justifying it (revive)

Is there a cleaner way to make sure we have "k8s.io/code-generator"? We don't need it at runtime, just when generating the deepcopy methods and the clientset for the custom CRD.

cognifloyd (commented):
Beyond adding tests, here are my open questions/TODOs.

runtime/kubernetes/build.go
```diff
@@ -39,22 +40,88 @@ func (c *client) InspectBuild(ctx context.Context, b *pipeline.Build) ([]byte, e
...
 	// https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1?tab=doc#ObjectMeta
 	c.Pod.ObjectMeta = metav1.ObjectMeta{
-		Name:   b.ID,
-		Labels: map[string]string{"pipeline": b.ID},
+		Name:        b.ID,
+		Labels:      labels,
+		Annotations: c.PipelinePodTemplate.Meta.Annotations,
+	}
+
+	// TODO: Vela admin defined worker-specific: AutomountServiceAccountToken
```
cognifloyd, line 84: I did not include `AutomountServiceAccountToken` in this PR as it requires more thought. The use of this might need to change when I add a feature that I haven't filed an issue for yet (The k8s runtime will need to query the docker registry for images, which might mean it needs to do something different with the ServiceAccount and any ImagePullSecrets.)

runtime/kubernetes/container.go
```diff
@@ -172,7 +173,14 @@ func (c *client) SetupContainer(ctx context.Context, ctn *pipeline.Container) er
 		}
 	}
 
-	// TODO: add SecurityContext options (runAsUser, runAsNonRoot, sysctls)
+	if c.PipelinePodTemplate != nil && c.PipelinePodTemplate.Spec.Container != nil {
+		securityContext := c.PipelinePodTemplate.Spec.Container.SecurityContext
+
+		// TODO: add more SecurityContext options (runAsUser, runAsNonRoot, sysctls)
```
cognifloyd, line 179: I did not address these in this PR because:
- fixing the user issues is not straight forward:
    -  `runAsUser` is an int user id, but the `user` tag in the pipeline is a string, and
    - `runAsNonRoot` is a form of validation - it just causes containers to fail if the image is not configured to use root; it does not switch it to use a non-root user.
- `sysctls` requires more discussion and safety checking:
    - it might need some new tag(s) in the pipeline config and
    - this will need some kind of sanity-checking validation, perhaps only allowing a limited set of them. The kubernetes docs warn that only a subset of sysctls are namespaced sufficiently to use safely per-container, but otherwise should be handled at the node-level.

cognifloyd:
Rebased and squashed into logical commits again.

cognifloyd:
I finished writing tests for this - coverage increased (a bit).

cognifloyd:
This PR is not as scary as it looks by line numbers. Most of it comes from codegen.

Generated lines: +2119
Test lines: +568 , -23
Non-test lines: +490 , -17
**Total lines changed: +3177 , -40**

Line count break down
- Generated lines
    - +2119
        - +249 `runtime/kubernetes/apis/**/zz_*.go`
        - +722 `runtime/kubernetes/generated/**/*.go`
        - +1148 `runtime/kubernetes/generated/*.yaml`
- Test lines
    - +432 , -23 **go test lines**
        - +430 , -22 `runtime/kubernetes/*.go`
        - +2 , -1 `runtime/docker/*.go`
    - +136 **test fixtures**
        - +136 `runtime/kubernetes/testdata/*.yaml`
- Non-test lines
    - +390 , -17 **go code lines**
        - +2 , -2 `go.mod`
        - +4 `cmd/vela-worker/*.go`
        - +203 `runtime/kubernetes/apis/*.go`
        - +164 , -15 `runtime/kubernetes/*.go`
        - +17 `runtime/*.go`
    - +100 **codegen support**
        - +2 `.gitignore`
        - +81 `Makefile`
        - +14 `PROJECT`
        - +3 `runtime/kubernetes/codegen/header.go.txt`

cognifloyd:
I don't think squashing all of this into only one commit would be good.
I'm open to breaking this up into more PRs, but I need some ideas on where to do that.

cognifloyd:
I just installed this in my cluster and validated that settings from my CRD instance end up in the pipeline pods (tested w/ annotations, dnsConfig, and dnsPolicy).

kneal (approved):

kneal (approved):
LGTM 🐬

runtime/flags.go
```diff
@@ -36,6 +36,18 @@ var Flags = []cli.Flag{
 		Name:     "runtime.namespace",
 		Usage:    "namespace to use for the runtime (only used by kubernetes)",
 	},
+	&cli.StringFlag{
+		EnvVars:  []string{"VELA_RUNTIME_PODS_TEMPLATE_NAME", "RUNTIME_PODS_TEMPLATE_NAME"},
+		FilePath: "/vela/runtime/pods-template-name",
+		Name:     "runtime.pods-template-name",
```
wass3r, line 42: nit: i know we're not super consistent for the `Name` field here, but we do seem to settle on using `_` instead of `-` for `FilePath` values. But as `Name` is really only used internally, it's easier to change after the fact. `FilePath` not so much.

wass3r (changes requested):

runtime/flags.go
kneal: Yeah, I was honestly wondering if `pods` should be `runtime.pods` since `pods` have multiple things under it but idk you're right about the lack of consistency.

wass3r: true - we can add consistency to `Name` outside of this PR, but that should probably be a thing :)

cognifloyd: OK - so I think you're suggesting this change for FilePath:
```suggestion
		FilePath: "/vela/runtime/pods_template_name",
		Name:     "runtime.pods-template-name",
```

I don't understand the issue with the `Name` field however.
Are you suggesting we use `runtime.pods.template-name` and `runtime.pods.template-file`?

cognifloyd: I made the `FilePath` change.

wass3r (approved):

kneal (approved):
LGTM 🐬

runtime/kubernetes/container.go
cognifloyd: For future reference:

Also, I added `runAsNonRoot` for the pod SecurityContext, but I did not add in the container SecurityContext. We may need to add it at container-level as well when this becomes configurable from the pipeline.

`sysctls` is only in the pod-level SecurityContext, and I did implement it there. We may still need to add something to pipeline config and add safety checking around that.

Had I realized that when writing this comment, I would have dropped `sysctls` and possibly `runAsNonRoot` from this TODO.
:shrug:

PR #299, opened 2022-04-05
chore: Always make the built binary executable

copart-jafloyd:
I keep forgetting to make the binary executable after I build with `make build-static` and before I build my new docker image.
This PR should save me from myself.


cognifloyd:
Ooops. This is me - I just submitted it with the wrong account.

PR #304, opened 2022-04-14
refactor(executor): clarify why use different contexts

cognifloyd:
Closes https://github.com/go-vela/community/issues/562

~Pass the context along in the executor. The `buildCtx` is derived from the global ctx (`gctx`) so that signal-based worker shutdown trickles down to the `buildCtx` as well (see [cmd/vela-worker/start.go](https://github.com/go-vela/worker/blob/master/cmd/vela-worker/start.go#L27-L53)).~

I documented why we do not pass the executor context and derive the `buildCtx` from it.

Since there are several contexts in the `exec()` function, I renamed `ctx` to `timeoutCtx` to clarify which context is getting passed to each method.

### DIFF
--- a/cmd/vela-worker/exec.go
+++ b/cmd/vela-worker/exec.go
@@ -18,14 +18,14 @@ import (
 // exec is a helper function to poll the queue
 // and execute Vela pipelines for the Worker.
 // nolint: nilerr // ignore returning nil - don't want to crash worker
-func (w *Worker) exec(index int) error {
+func (w *Worker) exec(ctx context.Context, index int) error {
 	var err error
 
 	// setup the version
 	v := version.New()
 
 	// capture an item from the queue
-	item, err := w.Queue.Pop(context.Background())
+	item, err := w.Queue.Pop(ctx)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


cmd/vela-worker/exec.go
```diff
@@ -93,63 +93,58 @@ func (w *Worker) exec(index int) error {
...
 
-	logger.Info("destroying build")
-	// destroy the build with the executor
-	err = _executor.DestroyBuild(context.Background())
-	if err != nil {
-		logger.Errorf("unable to destroy build: %v", err)
-	}
-
-	logger.Info("completed build")
-
```
cognifloyd, line 153: I can submit https://github.com/go-vela/worker/pull/304/commits/50a9a40bd9c5d3ca03566fee3f54491010854f1a as a separate PR if needed.

cmd/vela-worker/exec.go
```diff
@@ -93,63 +93,58 @@ func (w *Worker) exec(index int) error {
...
-	ctx, timeout := context.WithTimeout(ctx, t)
+	timeoutCtx, timeout := context.WithTimeout(buildCtx, t)
 	defer timeout()
 
 	defer func() {
 		logger.Info("destroying build")
 		// destroy the build with the executor
-		err = _executor.DestroyBuild(context.Background())
+		// using buildCtx instead of timeoutCtx so that it happens after the timeout
+		err = _executor.DestroyBuild(buildCtx)
```
cognifloyd, line 109:
```suggestion
		err = _executor.DestroyBuild(context.Background())
```
@jbrockopp said that:
> I think we'll want to be careful about passing the same context around everywhere.
> For example, specific to DestroyBuild() , we intentionally pass context.Background() instead of the shared context.
> The reason for that is because we want to guarantee all build resources are destroyed. If you pass the shared context then you run the risk of the build not getting destroyed.

cmd/vela-worker/exec.go
```diff
@@ -93,63 +93,58 @@ func (w *Worker) exec(index int) error {
 		t = time.Duration(item.Repo.GetTimeout()) * time.Minute
 	}
 
-	// create a background context
-	ctx := context.Background()
+	// create a build context
+	buildCtx, buildDone := context.WithCancel(ctx)
+	defer buildDone()
```
cognifloyd, line 98: this change means that a worker shutdown is going to cancel the context instead of just killing things.

If we don't want to do this, then we should expand the comment to explain why not.

cognifloyd:
I'm reconsidering this PR, breaking off the commits that I'm sure about into separate PRs: #313 #314

cognifloyd:
I developed some of the ideas here into #317. Once that is merged, I think I can shrink this PR significantly into something that is at least more reviewable if not mergable.

cmd/vela-worker/exec.go
cognifloyd: OK. I reverted this so that DestroyBuild still gets a background context. I added a comment about why.

jbrockopp (changes requested):

jbrockopp (approved):
LGTM

PR #311, opened 2022-04-14
bugfix(k8s): avoid "index out of range" panic by replacing magic indexes with containerLookup map

cognifloyd:
A bug report in slack pointed out that with services the kubernetes Runtime sometimes gets the wrong container.
See: https://gophers.slack.com/archives/CNRRKE8KY/p1646859800708539?thread_ts=1646859573.285529&cid=CNRRKE8KY

> I am also seeing a crash when adding services to the pipeline, panic: runtime error: index out of range [-1].  It looks like the logic https://github.com/go-vela/worker/blob/7c9599fdd5b4cb9d26ff235d41479cfe56f4f3ba/runtime/kubernetes/image.go#L55-L57 to exclude the init isn’t holding true on my deployment when using services.

So, this replaces magic index calculation with an index lookup map.

### DIFF
--- a/runtime/kubernetes/build_test.go
+++ b/runtime/kubernetes/build_test.go
@@ -392,6 +392,7 @@ func TestKubernetes_AssembleBuild(t *testing.T) {
 	for _, test := range tests {
 		_engine, err := NewMock(test.k8sPod)
 		_engine.Pod = test.enginePod
+
 		_engine.containersLookup = map[string]int{}
 		for i, ctn := range test.enginePod.Spec.Containers {
 			_engine.containersLookup[ctn.Name] = i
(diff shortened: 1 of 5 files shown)


kneal (approved):
LGTM 🐬 

I think this is fine, my only concern reading it was ensuring we aren't depending on the order of items in the map but I don't see that as the case.

cognifloyd:
Agreed - it really should be treated as an unordered lookup map. Thus far I'm only accessing it by key.

wass3r (dismissed):

wass3r:
sorry, maybe merged out of order, but this now has conflicts :(

cognifloyd:
> sorry, maybe merged out of order, but this now has conflicts :(

Nah. I didn't have an order in mind.

I merged in master. :)

jbrockopp (approved):
LGTM

JordanSussman (approved):

wass3r (approved):

PR #312, opened 2022-04-15
tests: Convert runtime tests to subtests

cognifloyd:
I use Goland to run `go test`. At least with Goland, `testing.Run()` improves the output and test selection. And it makes it easier to see which subtest has failed.

As I develop and refactor, I would really like to see this everywhere. This starts with all tests under `runtime/`.
After this is merged I can prepare another one that does `executor/`.

### DIFF
--- a/runtime/context_test.go
+++ b/runtime/context_test.go
@@ -50,11 +50,13 @@ func TestRuntime_FromContext(t *testing.T) {
 
 	// run tests
 	for _, test := range tests {
-		got := FromContext(test.context)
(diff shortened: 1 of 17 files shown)


kneal (dismissed):
LGTM 🐬  😍

jbrockopp (dismissed):
LGTM

wass3r (approved):
thanks for this! i like this approach better as well.

jbrockopp (approved):
LGTM

PR #313, opened 2022-04-17
refactor: drop duplicate executor.DestroyBuild call

cognifloyd:
The duplicate DestroyBuild call causes a bunch of errors in the log about the build already being destroyed.

So, let's just drop the duplicate call.

Extracted from #304 so it can be reviewed separately.

Fixes https://github.com/go-vela/community/issues/204

### DIFF
--- a/cmd/vela-worker/exec.go
+++ b/cmd/vela-worker/exec.go
@@ -108,6 +108,8 @@ func (w *Worker) exec(index int) error {
 		if err != nil {
 			logger.Errorf("unable to destroy build: %v", err)
 		}
+
+		logger.Info("completed build")
 	}()
 
 	logger.Info("creating build")
@@ -141,15 +143,6 @@ func (w *Worker) exec(index int) error {
 		logger.Errorf("unable to execute build: %v", err)
 		return nil
 	}
-
-	logger.Info("destroying build")
-	// destroy the build with the executor
-	err = _executor.DestroyBuild(context.Background())
-	if err != nil {
-		logger.Errorf("unable to destroy build: %v", err)
-	}
-
-	logger.Info("completed build")
 
 	return nil
 }


wass3r:
This is one of those things that I feel was put in place for a reason and I'm not sure what it was. I do know that there is quite a bit of "noise" that seems redundant in the logs for destroying things that's already destroyed. If there is a reason to double up on this, we probably should throw it in the comments.

@jbrockopp @kneal either of you recall why we double up on the cleanup here?

cognifloyd:
There's an outstanding issue about this: https://github.com/go-vela/community/issues/204

jbrockopp:
@cognifloyd I was able to dig up the PR that added this functionality in (duplicate `DestroyBuild()`):

https://github.com/go-vela/worker/pull/64

Unfortunately, I neglected to mention any details directly in the PR about why we needed the duplicate call 😞 

Apologies I can't offer more help but I'm struggling to recall exactly why we have it duplicated.

To some extent, I could see it being possible that we needed the duplicate call then but no longer need it now? 🤷 

@wass3r perhaps you could see if there is a closed issue in the internal Vela backlog with more details?

kneal:
If memory serves me correctly there was an issue with resource cleanup on the worker. We noticed that something was preventing the process from cleaning up stale docker resources (Network was the biggest issue since it had a static limit that could be defined on the daemon) so the duplicate call was added as an extra guarantee to ensure the resources were being deleted. 

I honestly, don't know if that bug(feature) still exists in the Docker runtime but I would highly recommend it's something we test. I don't think we ever came to a root cause of why it was occurring. 😢 Just kinda need a quick patch to stabilize the production instances that were running.

cognifloyd:
#64 did make things more robust because it calls DestroyBuild in a defer func. So, that's already much more reliable than only calling it at the end of the function iff nothing else failed first.

This keeps the defer call, but drops the one at the end.

jbrockopp:
@cognifloyd @kneal @wass3r 

I wanted to call out https://github.com/go-vela/worker/pull/41 which originally moved the `DestroyBuild()` func into a `defer` call.

I mention that PR specifically because it did the same thing that we're doing with this PR 😅 

Then, we ended up opening https://github.com/go-vela/worker/pull/64 which added the duplicate `DestroyBuild()` call.

I'm good with trying to move forward with this change and addressing any potential issues that pop up after the fact.

Sounds we're all on the same page that we don't want the duplicate call so hopefully we don't need it anymore 😃

kneal (approved):
LGTM 🐬

jbrockopp (approved):
LGTM

wass3r (approved):
Works for me. I didn't see anything internally to add any more details.

PR #314, opened 2022-04-17
refactor: cleanup context passing in kubernetes runtime

cognifloyd:
The API calls in the kubernetes runtime all use `context.Background()`. They should use the passed context instead.

Extracted from #304 so it can be reviewed separately.


jbrockopp (approved):
LGTM

wass3r (approved):

PR #316, opened 2022-04-18
tests: Convert internal/ tests to subtests

cognifloyd:
I use Goland to run `go test`. At least with Goland, `testing.Run()` improves the output and test selection. And it makes it easier to see which subtest has failed.

As I develop and refactor, I would really like to see this everywhere. This converts all tests under `internal/` like #312 did for `runtime/`.


jbrockopp (approved):
LGTM

wass3r (approved):

PR #317, opened 2022-04-20
enhancement: Make context handling for log streaming more explicit

cognifloyd:
Part of https://github.com/go-vela/community/issues/562

Implements the idea outlined in https://github.com/go-vela/community/issues/562#issuecomment-1100782504

### DIFF
--- a/executor/linux/build.go
+++ b/executor/linux/build.go
@@ -191,7 +191,7 @@ func (c *client) PlanBuild(ctx context.Context) error {
 // AssembleBuild prepares the containers within a build for execution.
 //
 // nolint: funlen // ignore function length due to comments and logging messages
-func (c *client) AssembleBuild(ctx context.Context) error {
+func (c *client) AssembleBuild(ctx context.Context, streamRequests chan<- executor.StreamRequest) error {
 	// defer taking a snapshot of the build
 	//
 	// https://pkg.go.dev/github.com/go-vela/worker/internal/build#Snapshot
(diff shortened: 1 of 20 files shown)


internal/message/stream.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/go-vela/types/pipeline"
+)
+
+// StreamFunc is either StreamService or StreamStep in executor.Engine.
+type StreamFunc = func(context.Context, *pipeline.Container) error
+
+// StreamRequest is the message used to begin streaming for a container
+// (requests goes from ExecService / ExecStep to StreamBuild in executor).
+type StreamRequest struct {
```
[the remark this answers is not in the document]
cognifloyd: I pulled it out of the interface, so now this is entirely internal.

kneal (dismissed):
LGTM 🐬

cognifloyd:
I renamed `logCtx` to `streamCtx` because I will probably use `StreamBuild` to stream events as well as logs in the Kubernetes runtime (in a future PR).

I did the rename now since I was also merging in master, and that'll reduce the diff for the next PR without changing the diff/functionality of this one.

kneal (dismissed):
LGTM 🐬

executor/linux/build.go
```diff
@@ -491,6 +491,50 @@ func (c *client) ExecBuild(ctx context.Context) error {
...
+
+	defer func() {
+		c.Logger.Trace("waiting for stream functions to return")
+
+		err := streams.Wait()
+		if err != nil {
+			c.Logger.Errorf("error in a stream request, %v", err)
+		}
+
+		c.Logger.Trace("all stream functions have returned")
```
cognifloyd, line 510: I'm debating. Should this be a different level like Info? That way you can see that each of the stream goroutines have returned in prod.
```suggestion
		c.Logger.Info("all stream functions have returned")
```

wass3r: after running a few test runs, i agree. that is nice confirmation.

wass3r (commented):
ltgm as well - thanks for all the work! if you want to make that one log level change, i think we're good.

i do see some inconsistencies with logging mentioned somewhere else too, sometimes using .WithFields and duplicating info in the actual message and sometimes not, but that's outside of the scope of this PR.

cognifloyd:
@wass3r Fixed the log entry. :) This should be ready when you are.

PR #308, opened 2022-04-14
bugfix(k8s): avoid InspectContainer panic where container is not Terminated

cognifloyd:
Closes https://github.com/go-vela/community/issues/564

slightly refactors the test so that I could include a failing test case.

PR #323, opened 2022-04-25
bugfix(k8s): do not log false-positive error when pause image is running

cognifloyd:
Follow up for #308 which successfully prevented a panic, but was still logging errors (about the pause image running) when it shouldn't have. It seems the test case added in #308 did not cover the pause image running, so this adds a couple of tests for that to ensure that that does NOT fail.

We have to use the canonical (parsed) pause image or it won't match at runtime:
https://github.com/go-vela/worker/blob/c25339835ea70c10e00d947ab3c08e2ab31b2135/runtime/kubernetes/container.go#L138

I also check for the simple non-canonical image name as well to simplify any unrelated tests that might incidentally use this.

~Also add a helpful note/comment about ContainerState objects.~

### DIFF
--- a/runtime/kubernetes/container.go
+++ b/runtime/kubernetes/container.go
@@ -53,17 +53,9 @@ func (c *client) InspectContainer(ctx context.Context, ctn *pipeline.Container) 
 
 		// avoid a panic if the build ends without terminating all containers
 		if cst.State.Terminated == nil {
-			for _, container := range pod.Spec.Containers {
-				if cst.Name != container.Name {
-					continue
-				}
-
-				// steps that were not executed will still be "running" the pause image as expected.
-				if container.Image == pauseImage {
-					return nil
-				}
-
-				break
+			// steps that were not executed will still be "running" the pause image as expected.
+			if cst.Image == pauseImage || cst.Image == image.Parse(pauseImage) {
+				return nil
 			}
 
 			return fmt.Errorf("expected container %s to be terminated, got %v", ctn.ID, cst.State)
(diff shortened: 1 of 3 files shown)


runtime/kubernetes/container.go
```diff
@@ -53,17 +53,9 @@ func (c *client) InspectContainer(ctx context.Context, ctn *pipeline.Container)
 
 		// avoid a panic if the build ends without terminating all containers
 		if cst.State.Terminated == nil {
-			for _, container := range pod.Spec.Containers {
```
cognifloyd, line 56: I realized that the image is already in `cst` so we don't need to look it up. That simplified this logic quite a bit.

runtime/kubernetes/container.go
```diff
@@ -378,6 +370,10 @@ func (c *client) WaitContainer(ctx context.Context, ctn *pipeline.Container) err
 				continue
 			}
 
+			// cst.State has details about the cst.Image's exit.
+			// cst.LastTerminationState has details about the kubernetes/pause image's exit.
+			// cst.RestartCount is 1 at exit due to switch from kubernetes/pause to final image.
```
jbrockopp, line 375: Should these be removed?

jbrockopp (changes requested):
LGTM - one qq:

runtime/kubernetes/container.go
jbrockopp: I don't think removing it is necessary since it has a purpose 👍 

More so, I wanted to double check that this was copied in during developmment and then left unintentionally

jbrockopp (dismissed):
LGTM

cognifloyd:
Oops. I missed that you approved. In any case, I dropped the comment, I can add it later if needed.

wass3r (approved):
ty

PR #326, opened 2022-04-26
feat(internal/skip): add event action handling for skip method

ecrupper:
Part of the effort for https://github.com/go-vela/community/issues/159

When we determine whether a container should skip at runtime, we check that the events match. With the introduction of `event_action` for scoping, we need to form that string in order to not skip those containers.


cognifloyd (approved):

wass3r (dismissed):

wass3r (changes requested):
can we throw a test into skip_test to cover action?

wass3r (dismissed):
Thanks

cognifloyd (approved):

PR #329, opened 2022-04-28
bugfix: make sure kubernetes log streaming gets canceled with build

cognifloyd:
While implementing #304 (improving context usage) and then #317 (improve worker log context management) I noticed that the Kubernetes Runtime's `TailContainer` ignores the context as it makes its requests. This fixes that so that tailing will stop if the streamCtx (which gets passed to TailContainer) is canceled.

### DIFF
--- a/runtime/kubernetes/container.go
+++ b/runtime/kubernetes/container.go
@@ -238,6 +238,10 @@ func (c *client) setupContainerEnvironment(ctn *pipeline.Container) error {
 func (c *client) TailContainer(ctx context.Context, ctn *pipeline.Container) (io.ReadCloser, error) {
 	c.Logger.Tracef("tailing output for container %s", ctn.ID)
 
+	// create a logsContext that will be canceled at the end of this
+	logsContext, logsDone := context.WithCancel(ctx)
+	defer logsDone()
+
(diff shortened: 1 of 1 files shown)


wass3r (approved):

jbrockopp (approved):
LGTM

runtime/kubernetes/container.go
```diff
@@ -238,6 +238,10 @@ func (c *client) setupContainerEnvironment(ctn *pipeline.Container) error {
 func (c *client) TailContainer(ctx context.Context, ctn *pipeline.Container) (io.ReadCloser, error) {
 	c.Logger.Tracef("tailing output for container %s", ctn.ID)
 
+	// create a logsContext that will be canceled at the end of this
+	logsContext, logsDone := context.WithCancel(ctx)
+	defer logsDone()
+
```
cognifloyd, line 244: Aargh. This is wrong.

We can't create a context here because we expect (at least part of) logsFunc to continue running after TailContainer returns. Then StreamStep/StreamService/etc read from the returned ReadCloser until the end of the stream. If the context gets canceled at the end of TailContainer, then the streaming is stopped before Stream* has read anything from it.

PR #331, opened 2022-04-30
feat(k8s): Make RunContainer detect/report more error conditions like image pull errors

cognifloyd:
This is an alternate implementation of #279 using the `SharedInformer` infra to watch events introduced in #519.
Closes #279

## Overview

The Docker Runtime surfaces image pull errors naturally because of the blocking synchronous requests to pull an image or start a container.

The Kubernetes Runtime has to watch for events that show any errors with pulling images.

## `RunContainer` changes in k8s runtime

Wait until one of these conditions before returning:
- build canceled
- container is running
- got an image pull error

## `WaitContainer` changes in k8s runtime

Wait until one of these conditions before returning:
- build canceled (new)
- container is terminated (was already doing this)

## `containerTracker` additions

I added signal channels for `Running` and `ImagePulled`. I also added a channel for receiving image pull errors in `RunContainer`.

## `podTracker` changes

I need pod `Name` and `Namespace` in more places, so I copy those into the podTracker now.

I was surprised that the `SharedInformerFactory` I used to create the Informer/Lister for watching the build Pod did NOT work for watching events. Apparently, events don't get labeled, so I had to add a separate `eventInformerFactory` that uses a FieldSelector instead of a LabelSelector when making the list/watch API calls.

I did not rename `informerFactory` to `podInformerFactory` because it can easily be used for other k8s resources in the future (if needed), just not for Events.

I also added the Event event handler funcs (wow that was a mouth full - events about Events).

`podTracker.inspectContainerEvent()` (called by the Event event handlers) is responsible for
- sending the image pull events to `RunContainer`
- closing the `ImagePulled` signal channel when an event indicates that the pull was successful

I also adjusted `podTracker.inspectContainerStatuses()` so that it can also:
- surface image pull errors (in case the error is visible there, but not through the Events API - which is unlikely, but there doesn't seem to be any kind of guarantees about event availability, so this covers our bases)
- close the `Running` signal channel when the container status shows that it is running.

TODO:
- [x] merge #302
- [x] rebase
- [ ] write tests

### DIFF
--- a/runtime/kubernetes/container.go
+++ b/runtime/kubernetes/container.go
@@ -81,6 +81,12 @@ func (c *client) RunContainer(ctx context.Context, ctn *pipeline.Container, b *p
 		return err
 	}
 
+	// get the containerTracker for this container
+	ctnTracker, ok := c.PodTracker.Containers[ctn.ID]
+	if !ok {
+		return fmt.Errorf("containerTracker missing for %s", ctn.ID)
+	}
+
 	// set the pod container image to the parsed step image
 	c.Pod.Spec.Containers[c.containersLookup[ctn.ID]].Image = _image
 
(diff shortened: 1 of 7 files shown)


runtime/kubernetes/container.go
```diff
@@ -23,6 +23,38 @@ import (
...
+	// nolint: godot // commented code is not a sentence
+	// known scheduler event reasons can be found here:
+	// https://github.com/kubernetes/kubernetes/blob/v1.23.6/pkg/scheduler/schedule_one.go
+	//reasonFailedScheduling = "FailedScheduling"
+	//reasonScheduled        = "Scheduled"
+
+	// known kubelet event reasons are listed here:
+	// https://github.com/kubernetes/kubernetes/blob/v1.23.6/pkg/kubelet/events/event.go
+
+	// kubelet image event reasons.
```
kneal, line 37: Would it make sense to just use their constants or wrap them at least so we know when they change?

runtime/kubernetes/pod_tracker.go
```diff
@@ -143,6 +170,83 @@ func (p *podTracker) getTrackedPod(obj interface{}) *v1.Pod {
...
+		"handling %s event update event for %s: fieldPath=%v",
+		newEvent.Type, // Normal, Warning
+		p.TrackedPod,
+		newEvent.InvolvedObject.FieldPath,
+	)
+
+	p.inspectContainerEvent(newEvent)
+}
+
+// HandleEventDelete is an DeleteFunc for cache.ResourceEventHandlerFuncs for Events.
```
kneal, line 211: If this is dead code, I'd prefer we delete it.

kneal (changes requested):

runtime/kubernetes/pod_tracker.go
cognifloyd: deleted

runtime/kubernetes/container.go
kneal: Cool, just checking. I just noticed when I went to the code they were public on the package. So, thought they might be importable.

kneal (approved):
LGTM 🐬

runtime/kubernetes/container.go
jbrockopp: Thought I'd provide a bit of clarity on this subject...

The tl;dr is you can vendor from `k8s.io/kubernetes` but its not pretty 😅 

Here's the reason why `go get k8s.io/kubernetes` fails:

https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505607224

And to actually vendor from `k8s.io/kubernetes`, you have to add a `replace` directive for all nested packages:

https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505627280

If we want, there are people who've scripted the approach to this so updating the version is easier:

https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597

To see a real world example of what this looks like in the `go.mod` file:

https://github.com/kubernetes-sigs/cloud-provider-huaweicloud/blob/9589bc854c29e399476479f9c5aa9599b2064da5/go.mod#L18-L40

jbrockopp (approved):
LGTM

runtime/kubernetes/container.go
cognifloyd: Eww... All that work just to reuse some constants... Not worth it imo.

jbrockopp: Agreed 😄

cognifloyd:
I'm running into some issues with this, so I'm moving it back to draft for now.

PR #336, opened 2022-05-09
tests: add integration test for exec()

cognifloyd:
Part of https://github.com/go-vela/community/issues/585

This adds some integration tests for the exec() function. It only adds them for the docker runtime, as it is currently hanging for kubernetes. I will fix the kubernetes runtime in follow-up PRs.

I also had to add a mock API server for the worker. I based it on the server's mock.


mock/worker/server.go
```diff
@@ -0,0 +1,41 @@
...
+func FakeHandler() http.Handler {
+	gin.SetMode(gin.TestMode)
+
+	e := gin.New()
+
+	// mock endpoints for utility calls
+	//e.GET("/health", getHealth)
+	//e.GET("/metrics", getMetrics)
+	//e.GET("/version", getVersion)
+	//e.POST("/api/v1/shutdown", postShutdown)
```
cognifloyd, line 24: I didn't add these, but we could.

PR #337, opened 2022-05-10
bugfix(k8s): Drop TailContainer's logsContext to avoid early cancel

cognifloyd:
Partially reverts #329 (see https://github.com/go-vela/worker/pull/329#discussion_r869658946).
In juggling all of my PRs, I thought I tested them merged together, but somehow I missed the breakage this caused.

We expect (at least part of) logsFunc to continue running after TailContainer returns.
Then StreamStep/StreamService/etc read from the returned ReadCloser until the end of the stream.

So, if the context gets canceled at the end of TailContainer (as introduced in #329), then the streaming is stopped before `Stream*` has read anything from it. So, log capture in the kubernetes runtime is currently broken for v0.14 😢 .

This fixes it by not creating+cancelling a new context in kubernetes TailContainer.

Also, I added/improved a log message about streaming output in each of the linux and local executors.

### DIFF
--- a/executor/linux/build.go
+++ b/executor/linux/build.go
@@ -536,6 +536,7 @@ func (c *client) StreamBuild(ctx context.Context) error {
 				return nil
 			})
 		case <-ctx.Done():
+			c.Logger.Debug("streaming context canceled")
 			// build done or canceled
 			return nil
 		}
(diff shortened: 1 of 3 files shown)


wass3r (approved):

jbrockopp (approved):
LGTM

PR #340, opened 2022-05-12
Allow creating mocks via `runtime.New()`

cognifloyd:
While writing `exec()` tests in #336, I found that I didn't have a good way to make sure the runtime was using a mock.

So, this adds Mock to the runtime setup which swaps out `docker.New()` for `docker.NewMock()` and swaps out `kubernetes.New()` for `kubernetes.NewMock()`.

It uses the `Mock` var introduced in #339, so that needs to be merged first. Once merged, I'll rebase this and mark it ready for review.


jbrockopp (approved):
LGTM

wass3r (approved):

PR #359, opened 2022-08-01
feat(executor/step): utilize trusted field to gate-keep privileged flag

ecrupper:
Currently there isn't a use for the `trusted` field for the repo resource. Given the potential security problems of the `privileged` flag, I think this is a perfect opportunity to utilize the field to restrict access to privileged containers. 

Docs: https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities

### DIFF
--- a/executor/linux/step.go
+++ b/executor/linux/step.go
@@ -82,6 +82,13 @@ func (c *client) PlanStep(ctx context.Context, ctn *pipeline.Container) error {
 	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithField
 	logger := c.Logger.WithField("step", ctn.Name)
 
+	if ctn.Privileged && !c.repo.GetTrusted() {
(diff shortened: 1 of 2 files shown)


cognifloyd:
Cool. Should this go in local executor as well? And does the UI/CLI already expose the trusted flag?

ecrupper:
> Cool. Should this go in local executor as well? And does the UI/CLI already expose the trusted flag?

My thinking was that people should be able to spin up their own privileged containers since it won't share the same engine as our workers. However, I can see wanting complete parity between local executor and linux. 

As of now, the UI does not expose it and does not allow updating, and the CLI shows it when running like `view repo` but does not allow editing it because our API ignores any attempts to do so.

ecrupper:
Closing this as the `privileged` flag doesn't really work since most worker set ups have an allow list for privileged images

PR #360, opened 2022-08-03
Specify Entrypoint correctly for Kubernetes

cognifloyd:
Docker ENTRYPOINT == K8s Container Command
Docker CMD == K8s Container Args

So, this uses Command instead of Args so that pipelines can actually
override the ENTRYPOINT specified in plugins.

### DIFF
--- a/runtime/kubernetes/container.go
+++ b/runtime/kubernetes/container.go
@@ -184,14 +184,14 @@ func (c *client) SetupContainer(ctx context.Context, ctn *pipeline.Container) er
 
 	// check if the entrypoint is provided
 	if len(ctn.Entrypoint) > 0 {
-		// add entrypoint to container config
-		container.Args = ctn.Entrypoint
+		// add docker ENTRYPOINT (aka k8s Command) to container config
+		container.Command = ctn.Entrypoint
 	}
 
 	// check if the commands are provided
 	if len(ctn.Commands) > 0 {
-		// add commands to container config
-		container.Args = append(container.Args, ctn.Commands...)
+		// add docker CMD (aka k8s Args) to container config
+		container.Args = ctn.Commands
 	}
 
 	// record the index for this container


jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

cognifloyd:
I've run into an issue with this I'm debugging. Do not merge yet

cognifloyd:
Dang it. This simple fix was too good to be true.
Entrypoint already gets overridden in the compiler:
https://github.com/go-vela/server/blob/f0210057c2d5aff435885fe31096ec57d1a6e923/compiler/native/script.go#L54-L58
```go
		// set the entrypoint for the step
		step.Entrypoint = []string{"/bin/sh", "-c"}

		// set the commands for the step
		step.Commands = []string{"echo $VELA_BUILD_SCRIPT | base64 -d | /bin/sh -e"}
```

But,
- the Kubernetes runtime relies on `docker.io/kubernetes/pause:latest`
- the `kubernetes/pause` container which does not have `/bin/sh`; it only has the `/pause` binary.
- all steps start off with `kubernetes/pause` until the step image gets patched in when the step should start
- In kubernetes, both Container.Command (aka docker ENTRYPOINT) and Container.Args (aka docker CMD) are immutable after pod creation
- So, the same entrypoint MUST work for both `kubernetes/pause` AND any step or service containers.

So, before any steps can run, there every step has this error in the logs (sadly only visible in debug logs):
```
failed to create containerd task: failed to create shim: OCI runtime create failed:
unable to start container process: exec: "/bin/sh": stat /bin/sh: no such file or directory: unknown
```

This fix, then, does not work as intended.

Implications:
Plugins, or at least the vela-git plugin, use `ENTRYPOINT` in the Dockerfile:
https://github.com/go-vela/vela-git/blob/main/Dockerfile#L25
```
ENTRYPOINT [ "/bin/vela-git" ]
```

With the Kubernetes runtime, we cannot effectively replace the docker ENTRYPOINT (set Container.Command). So, the vela plugins CANNOT be reused, which is apparently very common with the docker runtime.

I'm very tempted to say that we need to make the Kubernetes runtime use ephemeral containers for each step. But that's going to be a huge change. So, I'm going to close this as I don't see a simple way forward.

PR #368, opened 2022-09-07
feat(start/server): add minimum TLS version of 1.2 with option to set it differently

ecrupper:
`ListenAndServeTLS` defaults to TLS version 1.0 when a TLSConfig is not supplied. This can be a security issue due to 1.0's deprecated status. I've added a default minimum version of 1.2 but made it configurable. 

This PR also brings into question whether we want to enforce certain crypto cyphers for our certs. That would be a very similar process.


cmd/vela-worker/flags.go
```diff
@@ -85,6 +85,11 @@ func flags() []cli.Flag {
 			Name:    "server.cert-key",
 			Usage:   "optional TLS certificate key",
 		},
+		&cli.StringFlag{
+			EnvVars: []string{"WORKER_SERVER_TLS_MIN_VERSION", "VELA_SERVER_TLS_MIN_VERSION", "SERVER_TLS_MIN_VERSION"},
+			Name:    "server.tls-min-version",
+			Usage:   "optional TLS minimum version requirement",
+		},
```
wass3r, line 92: any reason to not set the default here where we define the flag like we do with other values?

cmd/vela-worker/server.go
```diff
@@ -56,10 +57,35 @@ func (w *Worker) server() (http.Handler, bool) {
 			logrus.Fatal("unable to run with TLS: No certificate provided")
 		}
 
-		return _server, true
+		// define TLS config struct for server start up
+		tlsCfg := &tls.Config{
+			MinVersion: tls.VersionTLS12,
```
wass3r, line 62: now with the default set where the flags are defined, this feels unnecessary. thoughts?

wass3r (approved):

plyr4 (approved):

wass3r (approved):

PR #376, opened 2022-09-14
fix(runtime/k8s/build): lowercase pod name for build

ecrupper:
Closes https://github.com/go-vela/community/issues/679

Kubernetes does not like any uppercase in their pod names, which causes any build ran from a repo with an uppercase letter in its name or org to fail.

### DIFF
--- a/runtime/kubernetes/build.go
+++ b/runtime/kubernetes/build.go
@@ -7,6 +7,7 @@ package kubernetes
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-vela/types/pipeline"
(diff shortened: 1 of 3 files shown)


wass3r (commented):
i wonder if this should be more comprehensive. @cognifloyd did this for container names: https://github.com/go-vela/types/blob/ef95941428d9195503c3db067a9f27447b4dcc16/pipeline/container.go#L272-L302, for reference. there are a few criteria for the spec. granted, there are a subset of these that would actually be allowed in org/repo names depending on SCM.

ecrupper:
> i wonder if this should be more comprehensive. @cognifloyd did this for container names: https://github.com/go-vela/types/blob/ef95941428d9195503c3db067a9f27447b4dcc16/pipeline/container.go#L272-L302, for reference. there are a few criteria for the spec. granted, there are a subset of these that would actually be allowed in org/repo names depending on SCM.

Yeah.. perhaps https://github.com/go-vela/types/blob/ef95941428d9195503c3db067a9f27447b4dcc16/pipeline/build.go#L112 would be a better place to start.

ecrupper:
Closing in favor of moving this change to types and being more comprehensive

PR #386, opened 2022-10-07
fix(actions): make sure to use latest Go version

wass3r:
same as https://github.com/go-vela/types/pull/270

also looks like we missed setting up Go for one of the jobs 😞

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,7 @@ jobs:
         # use version from go.mod file
         go-version-file: 'go.mod'
         cache: true
+        check-latest: true
 
     - name: build
       run: |
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -25,6 +25,7 @@ jobs:
         # use version from go.mod file
         go-version-file: 'go.mod'
         cache: true
+        check-latest: true
 
     - name: setup
       run: |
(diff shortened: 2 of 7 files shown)


plyr4 (approved):

plyr4:
good catch on the reviewdog step, i think that was my mistake :)

jbrockopp (changes requested):
AFAIK, setting up Go is not required when using `reviewdog/action-golangci-lint@v2`:

https://github.com/reviewdog/action-golangci-lint#migrating-from-v1-to-v2

> If your workflow have steps for setting up Go and caching go modules, they are no longer needed. reviewdog/action-golangci-lint@v2 now set up Go and cache modules automatically, so remove these steps.

wass3r:
interesting - thanks. wonder what kind of guarantee we have that it matches what we use for build. i'll have to go through all the PRs i opened and remove it from existing, affected workflows.

jbrockopp:
@wass3r Yeah, no problem! Before you do that, do you think it's worth keeping that configuration in place?

On the one hand, I'm a fan of not adding extra configuration unless its required to solve a use-case.

On the other hand, like you mentioned, this will guarantee the version of Go used and I don't think it hurts anything.

wass3r:
we commented at the same time - yea, i think i would probably leave it in place based on what i'm seeing. if there is another convincing argument, i'm happy to adjust.

jbrockopp (approved):
LGTM

PR #390, opened 2022-10-12
fix: Allow log streaming to take longer than build execution

cognifloyd:
With the kubernetes runtime, logs are not getting captured.

I've gone through a bunch of PRs to resolve this, all of which were good, but the issue still persists.
- #277
- #304
- #314
- #317
- #329
- #337

Sometimes the logs were captured but unreliably. Now, they are consistently not getting captured with the kubernetes runtime.

@JordanSussman, @jbrockopp and I started writing tests for the executor that use the kubernetes runtime (the full results of that effort are in the [executor-k8s-tests branch](https://github.com/go-vela/worker/tree/executor-k8s-tests)). Thanks to those tests, I was finally able to get a test (for `executor.StreamBuild`) that reproduced the issue, and subsequently developed the fix that is now in this PR.

So, this PR adds a new `executor.log_streaming_timeout` that only starts once the build finishes. After the build has finished, if the new timeout elapses, then the log streaming will also be canceled. The build process looks something like this:
- log streaming starts
- build starts
- build executes
- build finishes or times out (build context canceled)
- log streaming timer starts (once the build context is canceled)
- log streaming finishes or times out
- remaining go routines get cleaned up

So, this is a new feature that fixes a long standing bug that only presents itself with the kubernetes runtime.
It is helpful to review each commit separately.


JordanSussman (dismissed):

jbrockopp (approved):
LGTM

cognifloyd:
@wass3r you asked if the worker can pull another build before streaming has finished. Reviewing the code for `exec()`, I see I misspoke.

Today, streaming happens in a go routine where the only synchronization is cancelling the `buildCtx`; nothing else makes sure it exits.

https://github.com/go-vela/worker/blob/88acdef60f64fdb60e571127840f4f3579fd0d9a/cmd/vela-worker/exec.go#L140-L148

This PR does not change that. But, in a follow-up PR, I want to make sure destroying the build does not happen until streaming has finished. And in conjunction with that, we can make sure `exec()` does not return until streaming has finished which will prevent the worker from pulling a new build while streaming is going on.

cognifloyd:
Ok. I added a simple `sync.WaitGroup` in e882fd172a69ddd73612ba51e8cdd329c92f6951 to make sure StreamBuild finishes before we call DestroyBuild.

JordanSussman (approved):

jbrockopp (approved):
LGTM

plyr4 (approved):

wass3r (approved):
thanks for all the work on this!

we will need a separate PR to document the new flag.

PR #391, opened 2022-10-12
feat!: gate privileged images behind trusted repos

plyr4:
this PR introduces a feature flag for the worker executor: `VELA_EXECUTOR_ENFORCE_TRUSTED_REPOS`
the feature will utilize the `repo.trusted` field to gatekeep the execution of privileged images.
in other words with this feature enabled, only repos with `trusted==true` will be permitted to run pipelines containing the images supplied to the workers via the `RUNTIME_PRIVILEGED_IMAGES` variable. 

the default value for the flag `VELA_EXECUTOR_ENFORCE_TRUSTED_REPOS` is `true`, meaning this feature would be breaking with expected impact to current installations. ⚠️ 

the field `repo.trusted` already exists, but it isnt used by any part of the code, so its a relatively easy feature to implement.

essentially, before a build is setup it will be scanned for privileged images, and if it contains any then the repo will need to be trusted in order to proceed with execution.  
Screen Shot 2022-10-14 at 4 42 46 PM

### Edits

1. default value for `VELA_EXECUTOR_ENFORCE_TRUSTED_REPOS` was updated to `true` to provide a more secure default configuration


plyr4:
leaving in draft while I fix a few things
- step status should update to failure, it remains success right now
- tests

executor/linux/build.go
```diff
@@ -42,6 +44,76 @@ func (c *client) CreateBuild(ctx context.Context) error {
...
+			}
+
+			for _, pattern := range c.privilegedImages {
+				privileged, err := image.IsPrivilegedImage(container.Image, pattern)
+				if err != nil {
+					return fmt.Errorf("could not verify if image %s is privileged", container.Image)
+				}
+
+				if privileged {
+					containsPrivilegedImages = true
```
plyr4, line 77: could break on first "true"

ecrupper (approved):
lgtm, works great

executor/linux/build.go
```diff
@@ -42,6 +44,76 @@ func (c *client) CreateBuild(ctx context.Context) error {
...
+		containers = append(containers, c.pipeline.Services...)
+		for _, stage := range c.pipeline.Stages {
+			containers = append(containers, stage.Steps...)
+		}
+
+		for _, container := range containers {
+			// TODO: remove hardcoded reference
+			//
+			//nolint:goconst // ignore making a constant for now
+			if container.Name == "init" {
```
BrandonIngalls, line 66: Would the following config be exempt from the `containsPrivilegedImages` check, or does `container.Name` != the `name` key from the config?

```yaml
---
version: "1"

steps:
- name: init
  image: docker.io/library/alpine:3.16
  commands:
  - whoami
```

plyr4: good catch @BrandonIngalls. so that pipeline you pasted wouldnt work because of this line:
https://github.com/go-vela/worker/blob/main/executor/linux/step.go#L31-L33
which skips any steps and stages with "init" as the name from being created.

however a service named "init" would still get through so just to be safe i changed the code to skip over the hardcoded init image instead:
https://github.com/go-vela/worker/commit/784719552c2527a7ac6163716a80026a1ee82da6
https://github.com/go-vela/server/blob/main/compiler/native/initialize.go#L14

that should work better

ecrupper (dismissed):

originaldrock (approved):
LGTM

ecrupper (approved):
lgtm for the last time

PR #395, opened 2022-10-27
refactor(k8s): move test logic to MockKubernetesRuntime interface

cognifloyd:
When we add kubernetes-based tests to the executor(s) (see the [executor-k8s-tests](https://github.com/go-vela/worker/compare/executor-k8s-tests) branch), we will need to enough access to simulate certain kubernetes events. The `MockKubernetesRuntime` interface provides access to all of those mocking methods.

This PR only introduces the methods that can be immediately used in k8s tests. More mock utilities will be added in follow-ups.


jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

wass3r (approved):

PR #405, opened 2022-11-30
fix: move trusted repos check to handle dynamic images

plyr4:
using `pull: on_start` and dynamic image tags you could get the worker to crash when verifying pipelines that look like this:
```
version: "1"
environment:
  PRIV_IMAGE: "${VELA_BUILD_MESSAGE}"
  DYNAMIC_TAG: "${VELA_BUILD_NUMBER}"

steps:
  - name: run_priv
    pull: on_start
    image: 'docker.io/${PRIV_IMAGE}:${DYNAMIC_TAG}'
    commands:
      - "echo ${FOO_ARG}"
```

im simply moving the logic to below Create, where those variables would have been injected for steps and stages.

i also added secret origin containers to the check, skipped empty containers, and adjusted the tests.


KellyMerrick (approved):
lgtm

wass3r (approved):
liking the plentiful tests 🙇🏼 lgtm

PR #409, opened 2022-12-12
fix: revert log streaming commit

wass3r:
reverts https://github.com/go-vela/worker/pull/390/commits/e882fd172a69ddd73612ba51e8cdd329c92f6951 from https://github.com/go-vela/worker/pull/390

i acknowledge i was trying to account for an edge case, but it is true that a new build can be picked up while the stream log timeout timer is going. in fact, the timer gets kicked off regardless of whether there are logs still being streamed or not.

### DIFF
--- a/cmd/vela-worker/exec.go
+++ b/cmd/vela-worker/exec.go
@@ -6,7 +6,7 @@ package main
 
 import (
 	"context"
-	"sync"
+	// "sync"
 	"time"
 
 	"github.com/go-vela/worker/executor"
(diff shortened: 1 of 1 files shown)


plyr4 (approved):

timhuynh94 (approved):

cognifloyd (approved):
Yeah. I think we need something here, but we'll have to try something else. I'm still not sure why this code was hanging, but I'll keep investigating.

PR #410, opened 2022-12-13
fix: improve streaming coordination so it does not hang

cognifloyd:
It is helpful to review each commit of this PR.

It addresses the hanging we were experiencing because of #390 and reintroduces the commit reverted in #409. I tested in my cluster with this, and it successfully finishes one build and then pulls the next.

I found 3 problems with #390.
- StreamBuild could only stop waiting for new StreamRequests once the delayedCtx timeout elapsed.
- calling `cancelStreaming()` twice in StreamBuild was a little odd, so I added something to make it make sense.
- The order of defer statements in `operator.exec()` was sub-optimal, so I re-ordered a bit so that DestroyBuild gets called last.

### StreamBuild waiting too long for StreamRequests

This PR adds a new escape hatch: when the `streamRequests` channel gets closed (at the end of `ExecBuild`), then StreamBuild starts waiting for the stream functions to return. This should mean that the log streaming timer won't be triggered in more cases.

#### calling `cancelStreaming()` twice in StreamBuild

The only thing between calling `cancelStreaming()` in the StreamBuild goroutine that waits for functions, and the deferred  call of `cancelStreaming()` at the start of StreamBuild is the log entry that says all the functions returned. So, that wasn't providing much value. So, I added a little bit of value by blocking on the context that I just canceled. This way, more of the log messages from `WithDelayedCancelPropagation` will be logged before that "all functions have returned message". Sadly, there are still a couple of cases where those log messages might still print after, but it's a bit better.

#### order of defer statements in `operator.exec()`

Before this PR, the deferred statements were effectively called in this order:
- `DestroyBuild()`
- `timeout()`
- `done()`

That means that any goroutines that use the `buildCtx` or `timeoutCtx` to start cleaning up were not getting those signals until after the goroutine that calls `DestroyBuild` had started.

Now, they get called in this order:
- `timeout()`
- `done()`
- `DestroyBuild()`

So, everything should get the signal that it's time to cleanup before calling `DestroyBuild()`. This also means that the parent context that `StreamBuild` was waiting on will actually get the canceled signal (`ctx.Done()`) before `DestroyBuild`. So, the Streaming timer will actually start before the build gets destroyed.

When I added the WaitGroup with the old order, the `DestroyBuild()` goroutine was blocking, so `timeout()` and `done()` did not get called, so nothing got signaled that it was time to clean up.


KellyMerrick (approved):
lgtm

wass3r (approved):
thanks for working on this. lgtm

PR #418, opened 2023-01-21
refactor(executor tests): Make runtime a test arg

cognifloyd:

- renames var: `_runtime` to `_docker` (plus a few other related var renames)
- Adds `runtime` arg (to be accessed as `test.runtime`)
- In some tests, pass in the runtime itself in `test.runtime`.
- In other tests, pass a constant because the runtime will need to be recreated for every test.
- resorted some of the imports


wass3r (approved):
lgtm

plyr4 (approved):

PR #419, opened 2023-01-23
enhance(executor tests): test StreamBuild logging during build tests

cognifloyd:

- sorts the imports in `executor/linux/build_test.go`
- adjusts tests in `executor/linux/build_test.go` to check for logging un/expected errors. This is needed because logs are the only way to know if some areas of the code get hit, and because the logging happens in a goroutine, so we need to check the behavior of that goroutine when the function under test is doing its thing.


JordanSussman (approved):

jbrockopp (approved):
LGTM

plyr4 (approved):

PR #424, opened 2023-02-07
enhance(executor tests): Add test helpers to generate test Pods

cognifloyd:

- add several test helpers for use in generating the `pod` required for testing with the kubernetes runtime:
    - `func testPod(useStages bool) *v1.Pod`
    - `func testPodFor(build *pipeline.Build) *v1.Pod`
    - `func testContainerStatuses(...) []v1.ContainerStatus` _used by `testPod*`_
    - `func countBuildSteps(build *pipeline.Build) int` _used by `testContainerStatuses`_
- initialize the per-subtest k8s `_runtime` instances

For tests that initialize only one k8s runtime instance per test (vs per subtest), I will add that init logic when we add the test cases. Otherwise, the `_kubernetes` var would be flagged as unused.

### DIFF
--- a/executor/linux/build_test.go
+++ b/executor/linux/build_test.go
@@ -1090,6 +1090,12 @@ func TestLinux_PlanBuild(t *testing.T) {
 			var _runtime runtime.Engine
 
 			switch test.runtime {
+			case constants.DriverKubernetes:
+				_pod := testPodFor(_pipeline)
+				_runtime, err = kubernetes.NewMock(_pod)
+				if err != nil {
+					t.Errorf("unable to create kubernetes runtime engine: %v", err)
+				}
 			case constants.DriverDocker:
 				_runtime, err = docker.NewMock()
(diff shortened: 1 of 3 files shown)


executor/linux/build_test.go
```diff
@@ -1432,9 +1453,18 @@ func TestLinux_ExecBuild(t *testing.T) {
 			// Docker uses _ while Kubernetes uses -
 			_pipeline = _pipeline.Sanitize(test.runtime)
 
-			var _runtime runtime.Engine
+			var (
+				_runtime runtime.Engine
+				_pod     *v1.Pod
+			)
```
cognifloyd, line 1459: Unlike the other tests, this test will need access to _pod after initialize the k8s runtime.

jbrockopp (approved):
LGTM

ecrupper (approved):
LGTM

PR #425, opened 2023-02-21
enhance(executor tests): Call k8s SetupMock method after CreateBuild in tests

cognifloyd:
@JordanSussman, @jbrockopp and I worked on adding executor tests that use the kubernetes runtime. You can see the results of that effort in the [executor-k8s-tests](https://github.com/go-vela/worker/compare/executor-k8s-tests) branch. This PR prepares for getting all of those test cases added. I plan to add the test cases separately once more of the changes needed to manage the k8s mocks are included in the executor tests.

The Kubernetes runtime normally does some setup during AssembleBuild: We pre-populate the cache of the Mock PodTracker which watches the k8s API for changes to the pod (where build:pod maps 1:1). In any tests that call CreateBuild, we need to trigger that setup for the Mock after CreateBuild is called. So, we export a new `SetupMock` function that the executor tests can use to do this at the right point.

### DIFF
--- a/executor/linux/build_test.go
+++ b/executor/linux/build_test.go
@@ -1890,8 +1890,16 @@ func TestLinux_StreamBuild(t *testing.T) {
 				t.Errorf("%s unable to create build: %v", test.name, err)
 			}
 
-			// simulate ExecBuild() which runs concurrently with StreamBuild()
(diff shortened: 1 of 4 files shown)


wass3r (approved):
lgtm

jbrockopp (approved):
LGTM

plyr4 (approved):

PR #431, opened 2023-02-26
enhance(executor tests): Manage k8s mocks for Executor exec tests

cognifloyd:
@JordanSussman, @jbrockopp and I worked on adding executor tests that use the kubernetes runtime. You can see the results of that effort in the [executor-k8s-tests](https://github.com/go-vela/worker/compare/executor-k8s-tests) branch. This PR prepares for getting all of those test cases added. I plan to add the test cases separately once more of the changes needed to manage the k8s mocks are included in the executor tests.

This PR adds another part of the k8s mock management logic. In order to make the mocks during the ExecBuild test, we need 2 more exported methods in the kubernetes runtime:
- `StartPodTracker`
- `SimulateStatusUpdate`

K8s runtime tests have to manage starting the PodTracker which subscribes to a stream is k8s events. That is not part of the runtime interface, however, so we need to export a `StartPodTracker` method that the executor tests can use.

There are three kinds of k8s events that the k8s runtime watches for: changes to the pod itself, changes to the status of the pod, and generic k8s events. We use `SimulateResync` (introduced in a previous PR) to trigger events related to updating the pod itself. "Status" includes when the pod or a container is ready, running, terminated, etc. This is what `SimulateStatusUpdate` handles. I'm not using the third kind of generic k8s event yet in the kubernetes runtime. Once added, I don't think the executor tests will need to trigger that kind of event.

Finally we add calls to these and other mock methods in the ExecBuild test (to simulate execution) and the secret.exec tests. The StreamBuild test does not need to call `StartPodTracker` because that gets called as part of Runtime.StreamBuild. Instead, that test just needed a call to `MarkPodTrackerReady` (added in a previous PR) so that the runtime can make that call.

I also adjusted a few comments to fix a typo and suggest using the new mock helper methods.

### DIFF
--- a/runtime/kubernetes/build.go
+++ b/runtime/kubernetes/build.go
@@ -212,7 +212,7 @@ func (c *client) AssembleBuild(ctx context.Context, b *pipeline.Build) error {
 		}
 	}
 
-	// setup containerTeachers now that all containers are defined.
+	// setup containerTrackers now that all containers are defined.
 	c.PodTracker.TrackContainers(c.Pod.Spec.Containers)
 
 	// send signal to StreamBuild now that PodTracker is ready to be started.
--- a/executor/linux/build_test.go
+++ b/executor/linux/build_test.go
@@ -1570,7 +1570,9 @@ func TestLinux_ExecBuild(t *testing.T) {
 						if err != nil {
 							t.Errorf("%s - failed to simulate pod update: %s", test.name, err)
 						}
-						// TODO: maybe add a pause here
+
+						// simulate exec build duration
+						time.Sleep(100 * time.Microsecond)
 					}
 				}()
 			}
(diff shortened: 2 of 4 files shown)


runtime/kubernetes/build.go
```diff
@@ -212,7 +212,7 @@ func (c *client) AssembleBuild(ctx context.Context, b *pipeline.Build) error {
 		}
 	}
 
-	// setup containerTeachers now that all containers are defined.
+	// setup containerTrackers now that all containers are defined.
```
wass3r, line 215: oops :)

wass3r (approved):

JordanSussman (approved):

plyr4 (approved):

PR #437, opened 2023-02-28
chore: upgrade to docker v23 and add new mocked funcs

plyr4:


wass3rw3rk:
fixed by https://github.com/go-vela/worker/pull/535

PR #438, opened 2023-02-28
enhance(executor tests): Add kubernetes runtime test cases for Build tests

cognifloyd:
@JordanSussman, @jbrockopp and I worked on adding executor tests that use the kubernetes runtime. You can see the results of that effort in the [executor-k8s-tests](https://github.com/go-vela/worker/compare/executor-k8s-tests) branch.

This PR adds the test cases for `executor/linux/build_test.go` (except for `TestLinux_AssembleBuild_EnforceTrustedRepos` which was added after we did this work; we can add k8s runtime test cases to that test later).

All of these Build subtests use `runtime: constants.DriverKubernetes,` to switch between the runtime within each test. All of the logic to actually build the runtime instance was added in previous PRs allowing this PR to focus only on defining the test cases.

**Each subtest for k8s should be a copy of the docker subtest but with the k8s runtime constant instead.** The container IDs/names will also differ, as k8s uses `-` instead of `_` in the sanitized names.

### commented FIXME tests

Several of the test cases are commented out with this FIXME:
`// FIXME: make Kubernetes mock simulate failure similar to Docker mock`

These commented cases serve as a good indicator of places where the kubernetes runtime behavior might differ from the docker runtime behavior. For example, if the image is not found, that error is not currently captured in the k8s runtime, it just times out instead. This particular issue should be addressed by resolving https://github.com/go-vela/community/issues/579 (prob in https://github.com/go-vela/worker/pull/331). Once we have a way to capture those errors in the k8s runtime, then I will figure out how to adjust the k8s mock to appropriately fail where expected.


jbrockopp (approved):
LGTM

executor/linux/build_test.go
```diff
@@ -1797,6 +2089,29 @@ func TestLinux_StreamBuild(t *testing.T) {
...
+			pipeline:   "testdata/build/stages/basic.yml",
+			messageKey: "step",
+			streamFunc: func(c *client) message.StreamFunc {
+				return c.StreamStep
+			},
+			planFunc: func(c *client) planFuncType {
+				return c.PlanStep
+			},
+			ctn: &pipeline.Container{
+				ID:          "step-github-octocat-1-test-test",
```
plyr4, line 2106:
```suggestion
				ID:          "step-github-octocat-1-test",
```

plyr4 (approved):

executor/linux/build_test.go
cognifloyd: Stage test + step test

plyr4: oh! it seems so obvious now

PR #439, opened 2023-02-28
enhance(executor tests): Add kubernetes runtime test cases for Opts and Secrets tests

cognifloyd:
@JordanSussman, @jbrockopp and I worked on adding executor tests that use the kubernetes runtime. You can see the results of that effort in the [executor-k8s-tests](https://github.com/go-vela/worker/compare/executor-k8s-tests) branch.

This PR adds the test cases for `executor/linux/opts_test.go` and `executor/linux/secret_test.go`.

The subtests for `TestLinux_Secret_exec` uses `runtime: constants.DriverKubernetes,` to switch between the runtime within each test. For that test, the logic to actually build the runtime instance was added in previous PRs.

The rest of the tests only create one runtime for all the subtests (vs per subtest). These subtests use `runtime: _kubernetes` instead of passing in a constant. I did not initialize the k8s runtime in previous PRs to avoid adding vars like `_kubernetes` that are unused without the test cases that need them.

**Each subtest for k8s should be a copy of the docker subtest but with the k8s runtime (or the constant) instead.** The container IDs/names will also differ, as k8s uses `-` instead of `_` in the sanitized names.

### commented FIXME tests

Several of the test cases are commented out with this FIXME:
`// FIXME: make Kubernetes mock simulate failure similar to Docker mock`

These commented cases serve as a good indicator of places where the kubernetes runtime behavior might differ from the docker runtime behavior. For example, if the image is not found, that error is not currently captured in the k8s runtime, it just times out instead. This particular issue should be addressed by resolving https://github.com/go-vela/community/issues/579 (prob in https://github.com/go-vela/worker/pull/331). Once we have a way to capture those errors in the k8s runtime, then I will figure out how to adjust the k8s mock to appropriately fail where expected.


jbrockopp (approved):
LGTM

wass3r (approved):

PR #447, opened 2023-03-15
chore(log): remove byte-chunks log method

wass3r:
prior to this change, we provided two methods for uploading logs from the worker: byte-chunks and time-chunks. this change removes byte-chunks and makes time-chunks the default upload method. there are various issues with byte-chunks. 

one of the main problems is that byte-chunks has a high probability to thrash the database and affect its integrity, in particular for builds that are spewing logs at fast rates. the additional UPDATE calls in the database makes it prone to develop bloat quickly.

additional changes needed elsewhere:
- update docs to remove log method flag documentation, see https://github.com/go-vela/docs/blob/25dc9ffe9554ee59f351ae972b5fa53370a1d650/content/installation/worker/reference/executor.md?plain=1#L19
- update docker-compose files in core repos to remove env var, for example: https://github.com/go-vela/server/blob/a3ab6f02f60841a08ee12d695bf5a61311df310c/docker-compose.yml#L69


wass3r:
anyone not touching the defaults would be using `byte-chunks` and potentially incurring additional stress to their db/installation. the change would be an enhancement for those folks.

plyr4 (approved):

PR #449, opened 2023-03-17
feat: use validate-token endpoint in MustServer

plyr4:
this expands on https://github.com/go-vela/server/pull/790 to use the `/validate-token` endpoint to validate that incoming requests to the worker API are indeed coming from the server that has been configured to communicate with the workers on the behalf of the platform and users (build cancels).

since we are moving away from using "temp users" as the basis for server<->worker communication, this also deprecates `user.Establish` in favor of moving all server-secret validation into `MustServer`

depends on https://github.com/go-vela/server/pull/790 and https://github.com/go-vela/sdk-go/pull/212


router/middleware/perm/perm.go
```diff
@@ -19,19 +19,62 @@ import (
...
+				logrus.Error(err)
+			}
+
+			c.AbortWithStatusJSON(http.StatusInternalServerError, types.Error{Message: &msg})
 
-		err := c.Error(fmt.Errorf(msg))
+			return
+		}
+
+		vela, err := vela.NewClient(addr, "", nil)
```
plyr4, line 50: should we set the client ID to something like `vela-worker-$hostname`? or just `vela-worker`?

router/middleware/perm/perm.go
```diff
@@ -19,19 +19,62 @@ import (
...
+				logrus.Error(err)
+			}
+
+			c.AbortWithStatusJSON(http.StatusInternalServerError, types.Error{Message: &msg})
+
+			return
 		}
 
-		c.AbortWithStatusJSON(http.StatusUnauthorized, types.Error{Message: &msg})
+		vela.Authentication.SetTokenAuth(tkn)
```
plyr4, line 64: creating a client and using SetTokenAuth seemed like a safer approach than just hitting the API directly.
like, maybe token auth gets tweaked, or the client has new headers added, etc.

plyr4: i suppose we could pass around a shared client but it didnt seem as safe, for example with concurrent incoming cancel requests

router/middleware/perm/perm.go
```diff
@@ -19,19 +19,62 @@ import (
...
+
+		_, err = vela.Authentication.ValidateToken()
+		if err != nil {
```
plyr4, line 67: im almost positive if the new validate-token returns a 401 then it gets caught in this err but this was written in tandem so i still need to verify that. 
aka, should we also check the returned response's status code?

router/middleware/perm/perm.go
```diff
@@ -19,19 +19,62 @@ import (
...
+			msg := fmt.Sprintf("error validating token: %s", err)
+
+			err := c.Error(fmt.Errorf(msg))
+			if err != nil {
+				logrus.Error(err)
+			}
+
+			c.AbortWithStatusJSON(http.StatusInternalServerError, types.Error{Message: &msg})
```
plyr4, line 75: should we check the error returned by ValidateToken and return the same status code? 
aka, should this return a 401 when validate failed with a 401? and 500 when validate returned an actual error or a 500?

cognifloyd (dismissed):

router/middleware/perm/perm.go
```diff
@@ -9,29 +9,106 @@ import (
...
+			c.AbortWithStatusJSON(http.StatusBadRequest, err.Error())
 
-		if strings.EqualFold(u.GetName(), "vela-server") {
 			return
 		}
 
-		msg := fmt.Sprintf("User %s is not a platform admin", u.GetName())
+		// retrieve the configured server address from the context
+		addr, ok := c.MustGet("server-address").(string)
+		if !ok {
```
ecrupper, line 40: `c.MustGet` returns a panic if it doesn't find the value for the specified key, so the `ok` check will never actually happen AFAICT.

router/middleware/perm/perm.go
```diff
@@ -9,29 +9,106 @@ import (
...
+
+			err := c.Error(fmt.Errorf(msg))
+			if err != nil {
+				logrus.Error(err)
+			}
+
+			c.AbortWithStatusJSON(http.StatusInternalServerError, types.Error{Message: &msg})
+
+			return
 		}
```
ecrupper, line 66: Thoughts on changing this (and other error blocks like it) to:

```go
		if err != nil {
			retErr := fmt.Errorf("error creating vela client: %s", err)
		
			logrus.Error(retErr)
			
			c.AbortWithStatusJSON(http.StatusInternalServerError, retErr)

			return
		}
```
? A little more condensed, though I'm indifferent. Better than just logging it.

plyr4: also, should our pattern be to return the json wrapped error with `message:` or just the raw error as the string?

IE: `types.Error{Message: &msg}` vs `retErr` 

both patterns are used across multiple go-vela repos so its tough to say.

i can definitely go with `string` for simplicity

router/middleware/perm/perm.go
plyr4: good call im going to change that to `Get()`

cognifloyd (approved):

ecrupper (approved):

PR #452, opened 2023-03-29
enhance(auth): implement registration flow

ecrupper:
Ref: https://github.com/go-vela/community/blob/main/proposals/2023/03-15_worker-auth.md
Server Ref: https://github.com/go-vela/server/pull/790

This PR implements an optional worker registration flow and enhances the existing check in Go Routine to be more consequential. The security benefits are outlined in the proposal linked above, so I will focus more on this specific design.

To start, the primary structural change is the addition of these two fields, which will be mentioned often in the below explanations

```go
	Worker struct {
		Config        *Config
		Executors     map[int]executor.Engine
		Queue         queue.Service
		Runtime       runtime.Engine
		VelaClient    *vela.Client
+++		RegisterToken chan string 
+++		CheckedIn     bool
	}
```

#### Symmetric Token Flow
If the platform administrators supply a `VELA_SERVER_SECRET` in the environment as well as the corresponding `VELA_SECRET` in the server environment, then these PR changes will have very minimal impact on the worker.  The reason being is that the `RegisterToken` channel is filled immediately by the server-worker token, which does not expire. It is then transferred to the SDK client and does not get replaced for the duration of the worker container's existence.

The only area of impact for this setup is the dependency of queue pulling based on the success of the worker check in. This is where the `CheckedIn` field comes in to play. If, for whatever reason, the check in process fails, I would argue that we do not want this worker pulling jobs from the queue. There is a retry mechanism in place in order to protect against network blips as well as not severely inconvenience local dev (where the server is almost always still setting up when the worker container starts). 

#### Registration Flow
If the platform administrators do NOT supply the `VELA_SERVER_SECRET`, the worker will require an authenticated connection with the server before proceeding its operations. In short, it will launch its server routine as well as its signal routine, but the `phoning home` and `exec` routines will not run until a registration token fills the `RegisterToken` channel. 

This is where the new worker API endpoint `/register` comes into play. A platform admin will request a registration token using [this endpoint](https://github.com/go-vela/server/blob/8b1b8575e5398054058ab524b015efd5fe6a69ed/api/admin/worker.go#L48), which will then be used as the auth header for the worker's `/register` endpoint. The token will then be pushed to the channel, which has been added to the context, and the worker will know to resume operations (check-in + exec). 

From then on, the `phone home` routine will take care of maintaining the worker's authentication, exchanging worker auth tokens (which will have a lifespan of the check in interval plus a buffer), and placing the active token into the client's auth. 

For handling network blips and/or outages, the `RegisterToken` channel is again used to safeguard the worker from entering any forever-loop with a bad token while still adhering to a retry mechanism for short outages. The nitty gritty details of that are all in the code comments (please let me know if you'd like more clarity). 

Note: there are incoming CLI changes that will allow admins to request a registration token and hit that self-register worker endpoint in one fell swoop.

### DIFF
--- a/api/register.go
+++ b/api/register.go
@@ -26,6 +26,10 @@ import (
 //     description: Successfully passed token to worker
 //     schema:
 //       type: string
+//   '500':
+//     description: Unable to pass token to worker
+//     schema:
+//       "$ref": "#/definitions/Error"
 
 // Register will pass the token given in the request header to the register token
 // channel of the worker. This will unblock operation if the worker has not been


wass3rw3rk (approved):
nice work

plyr4 (approved):

PR #454, opened 2023-04-04
feat(runtime/docker): add drop kernel capabilities option to runtime flags

ecrupper:
This change allows platform administrators to launch their worker runtime with certain kernel capabilities dropped for each docker container.


plyr4:
awesome work, i really like that you made it configurable 💯

plyr4 (approved):

PR #461, opened 2023-04-14
enhance: operate/exec logging

plyr4:
just adding more logging around registration. also changed some casing to be more consistent with the other logs (lower-case).


wass3rw3rk (approved):

ecrupper (approved):
👍 let's go to the moon

PR #465, opened 2023-04-19
fix(register): verify hostname

wass3rw3rk:
this stuffs the worker hostname into the gin context which is used as a simple pre-validation that the incoming registration token is valid for the host. it improves the registration experience.

currently, if you supply a wrong hostname, the validation fails on the server AFTER successfully passing through the /register endpoint on the worker. as an admin, it looks like registration was successful when in fact, the worker is in a failing check in loop. this change should prevent that scenario.

### DIFF
--- a/api/register.go
+++ b/api/register.go
@@ -5,10 +5,12 @@
 package api
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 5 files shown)


ecrupper (approved):
lgtm nice

plyr4 (approved):

PR #472, opened 2023-05-08
enhance(exec): do not kill worker if build has been canceled and build token minting fails

ecrupper:
Using the status code for the response from build token minting, we can determine whether or not a build has been canceled while it was in the queue. If such a case arises, it is best to discard the item.

However, if the token minting failed due to poor authentication / problem with the server, it would make sense to kill the worker (and re-register, if that is enabled).

Dependent on: https://github.com/go-vela/server/pull/840
Part of the effort for: https://github.com/go-vela/community/issues/485

### DIFF
--- a/cmd/vela-worker/exec.go
+++ b/cmd/vela-worker/exec.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"context"
+	"net/http"
 	"sync"
 	"time"
 
@@ -19,7 +20,7 @@ import (
 // exec is a helper function to poll the queue
 // and execute Vela pipelines for the Worker.
 //
-//nolint:nilerr // ignore returning nil - don't want to crash worker
+//nolint:nilerr,funlen // ignore returning nil - don't want to crash worker
 func (w *Worker) exec(index int) error {
 	var err error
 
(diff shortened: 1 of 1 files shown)


cmd/vela-worker/exec.go
```diff
@@ -37,9 +38,16 @@ func (w *Worker) exec(index int) error {
 	}
 
 	// retrieve a build token from the server to setup the execBuildClient
-	bt, _, err := w.VelaClient.Build.GetBuildToken(item.Repo.GetOrg(), item.Repo.GetName(), item.Build.GetNumber())
+	bt, resp, err := w.VelaClient.Build.GetBuildToken(item.Repo.GetOrg(), item.Repo.GetName(), item.Build.GetNumber())
 	if err != nil {
 		logrus.Errorf("unable to retrieve build token: %s", err)
+
+		// build is not in pending state — user canceled build while it was in queue. Pop, discard, move on.
+		if resp != nil && resp.StatusCode == http.StatusBadRequest {
```
cognifloyd, line 46: Is there any other reason we might get a 400 error here?

cognifloyd: Yeah. If this returned a code that is unique to "the build is not runnable" (not in pending or running states) that would be great.

Maybe one of
- 409 Conflict https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409
- 410 Gone https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410

cognifloyd (dismissed):

plyr4 (approved):

KellyMerrick (approved):
lgtm

PR #478, opened 2023-05-18
feat!: Use queue Item.ItemVersion field to fail stale builds

cognifloyd:
Part of https://github.com/go-vela/community/issues/811
Depends on https://github.com/go-vela/types/pull/292

This should allow the worker to invalidate stale queued builds. A stale build is a build that was queued before a Vela upgrade or downgrade if the two Vela versions do not share the same queue ItemVersion.

In the future, we could do something even nicer and ask the server to recompile and requeue the build.

Closes https://github.com/go-vela/community/issues/811

### DIFF
--- a/cmd/vela-worker/exec.go
+++ b/cmd/vela-worker/exec.go
@@ -10,10 +10,11 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-vela/types"
+	"github.com/go-vela/types/constants"
 	"github.com/go-vela/worker/executor"
 	"github.com/go-vela/worker/runtime"
 	"github.com/go-vela/worker/version"
-
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/vela-worker/exec.go
```diff
@@ -70,6 +71,25 @@ func (w *Worker) exec(index int) error {
...
+	// handle stale item queued before a Vela upgrade or downgrade.
+	if item.ItemVersion != types.ItemVersion {
+		// If the ItemVersion is older or newer than what we expect, then it might
+		// not be safe to process the build. Fail the build and loop to the next item.
+		// TODO: Ask the server to re-compile and requeue the build instead of failing it.
+		logrus.Errorf("Failing stale queued build due to wrong item version: want %d, got %d", types.ItemVersion, item.ItemVersion)
+
+		build := item.Build
+		build.SetStatus(constants.StatusFailure)
+
```
wass3r, line 83: should we set a nice, user friendly message on the build also?

cognifloyd: Oh. Excellent idea. Done!

ecrupper (approved):

wass3r (approved):

PR #479, opened 2023-05-23
feat: add support for schedules

jbrockopp:
xref: https://github.com/go-vela/community/issues/538

Part of https://github.com/go-vela/community/pull/772

Dependent on https://github.com/go-vela/server/pull/836

This adds support for running pipelines with `event: schedule` in the ruleset.


ecrupper:
I see that this `Skip` check extends to other event types, but I'm sort of at a loss for how this ever gets called, since we purge compile-time rulesets here: https://github.com/go-vela/server/blob/5a9a92d279e058b75b37e589e1eef2ec63f550a6/compiler/native/transform.go#L197. AFAICT, the only use this `Skip` function serves is determining if the `status` is valid against the ruledata. 

Is there a case where `Purge` wouldn't have caught this?

plyr4 (approved):

jbrockopp:
@ecrupper `Purge()` does both filtering out steps based off `ruleset` as well as setting the `Number` field.

It is a compile-time filter so it doesn't account for `status` in a `ruleset` (intentionally).

`Skip()` is a build-time filter based off `ruleset` and does account for `status` in a `ruleset`.

They do some similar things but aren't an exact match since they have different purposes.

ecrupper:
@jbrockopp Yeah, I guess what I'm trying to say is perhaps there is an opportunity for `container.Execute` to not require all the compile-time ruledata to be set 🤷  — then this wouldn't be necessary any time we want to add a rule in the future.

ecrupper (approved):

JordanSussman (approved):

PR #510, opened 2023-09-19
enhance(init): display basic worker info in init step logs

ecrupper:
Closes https://github.com/go-vela/community/issues/125.

While we can leverage the hover display on the build nav bar to check the worker that ran a specific build, I've found this to be a bit finicky and more difficult to analyze older builds. Throwing some basic worker / runtime info in the init step seems like a fairly simple win. 

I can also foresee this change being temporary if/when [the init step overhaul proposal](https://github.com/go-vela/community/pull/789) gets off the ground and running.

### DIFF
--- a/executor/linux/build.go
+++ b/executor/linux/build.go
@@ -98,6 +98,9 @@ func (c *client) PlanBuild(ctx context.Context) error {
 		return err
 	}
 
+	// put worker information into init logs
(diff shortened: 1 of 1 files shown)


plyr4 (approved):
nice

wass3rw3rk (approved):

PR #517, opened 2023-10-02
fix(queue): avoid worker restart on queue fail

wass3rw3rk:
this should make it nicer for admins in the instance there is a temporary issue on the queue availability. previously, this could cause the worker to become unregistered resulting in admin work. this change will not error the worker (and cause it to restart). it will log the error and wait for the configured queue timeout before continuing business as usual.


plyr4 (approved):

ecrupper (approved):

KellyMerrick (approved):
lgtm

timhuynh94 (approved):

PR #522, opened 2023-10-06
fix(executor/local): do not override env with machine details for local exec

ecrupper:
This is the second of three changes that will enhance the `vela exec pipeline` command. These changes fall under the `Worker` section below.

### The Issue

`vela exec pipeline` does not work with templates. It also does not work with custom environments if the user is trying to emulate any of the default environment variables, such as `VELA_BUILD_COMMIT`.

### Server Changes

1. Execute the local environment sourcing _after_ the default env override. This will ensure that user-set values that would be found in the default environment will be properly interpreted by the compiler.
2. Adjustments to `getTemplate` that allows the compiler to read both from local file tree as well as remote using a GitHub token.

### Worker Changes

1. Remove calls to merge environment with default environment found in [executor/local/step.go](https://github.com/go-vela/worker/blob/be492bd1e78717a54717c183983845be3234c236/executor/local/step.go#L64-L70). This call is unnecessary and will only overwrite user-supplied environment values found in the default env.

### CLI Changes

1. Support for `--template`, `--template-file`, `--compiler.github.token`, and `--compiler.github.url` flags to the `exec pipeline` command. This will allow users the option to leverage templates in local pipeline execution, both local and remote.
2. Support for `--env-file` and `--env-file-path` to allow users to source environment variables from a file.


wass3rw3rk (dismissed):

wass3rw3rk (approved):

executor/local/stage.go
```diff
@@ -90,6 +91,7 @@ func (c *client) ExecStage(ctx context.Context, s *pipeline.Stage, m *sync.Map)
 		//
 		// https://pkg.go.dev/github.com/go-vela/worker/internal/step#Skip
 		if step.Skip(_step, c.build, c.repo) {
+			logrus.Infof("Skipping step %s due to ruleset", _step.Name)
```
plyr4, line 94: disclaimer i havent tested the PR to see what the logging actually looks like mid-execution. but we usually print to stdout during pipeline execution, should we do that here to follow the pattern?

ecrupper: It looks the same as all the `init` step logs: 
```sh
vela exec pipeline 
DEBU[0000] validating pipeline configuration            
DEBU[0000] Creating registry clients from CLI configuration 
...
[step: init] > Preparing service images...
[step: init] > Preparing stage images...
[step: init] > Preparing step images...
DEBU[0000] streaming build logs                         
[step: init] $ docker image inspect golang:latest
sha256:2159148dcc081245165b2aa99fc5a94ca9818bece66839d8eb11c9335ae9e688
...
[step: docker-build] INFO[0000] Skipping push to container registry due to --no-push flag 
INFO[0004] Skipping step notify_slack due to ruleset 
[step: random] > Streaming container 'step_EastonCrupper_myvela_1_random'...
```
Think that's decent enough?

wass3r (approved):

executor/local/stage.go
plyr4: ah sweet, the example is great. it views more like a log from the worker process controlling the execution flow, i like it!

plyr4 (approved):

PR #526, opened 2023-10-24
feat(executor/secrets): lazy loading secrets

claire1618:
Here's the issue: [](https://github.com/go-vela/community/issues/196)

The changes here make it so that secrets can be pulled in at the beginning of a build, or before each step, depending on what the user sets.


wass3r (changes requested):
Looks like there are a few things in here leftover from local debugging/testing?

executor/linux/build.go
```diff
@@ -164,7 +165,14 @@ func (c *client) PlanBuild(ctx context.Context) error {
 			continue
 		}
 
-		c.Logger.Infof("pulling %s %s secret %s", secret.Engine, secret.Type, secret.Name)
+		// only pull in secrets that are set to be pulled in at the start
+		if strings.EqualFold(secret.Pull, constants.SecretPullStep) {
+			c.Logger.Infof("SecretPullBuild")
```
plyr4, line 170: should we remove this debug line, or adjust it to make it look consistent with our other logs?

executor/linux/build.go
```diff
@@ -419,6 +427,7 @@ func (c *client) AssembleBuild(ctx context.Context) error {
 			_log.AppendData([]byte(fmt.Sprintf("Privileges verified for image %s\n", container.Image)))
 		}
 
+		// TODO
```
plyr4, line 430: is there something left to do here?

claire1618: nope! I removed it

executor/linux/build.go
claire1618: I adjusted it!

timhuynh94 (dismissed):

executor/linux/build.go
```diff
@@ -520,6 +529,61 @@ func (c *client) ExecBuild(ctx context.Context) error {
...
+					secret.Engine, secret.Type, s.GetOrg(), s.GetRepo(), s.GetName())))
+
+			sRaw, err := json.MarshalIndent(s.Sanitize(), "", " ")
+			if err != nil {
+				c.err = err
+				return fmt.Errorf("unable to decode secret: %w", err)
+			}
+
+			_log.AppendData([]byte(string(sRaw)[strings.Index(string(sRaw), "name")-2 : strings.Index(string(sRaw), "value")-2]))
+			_log.AppendData([]byte(string(sRaw)[strings.Index(string(sRaw), "images")-2 : strings.Index(string(sRaw), "created_at")-2]))
```
plyr4, line 570: can we add some index bounds checks here?

executor/linux/build.go
wass3rw3rk: what was the adjustment? i don't think this line was adjusted.

claire1618: it should be there now, something weird was going on but I fixed it

executor/linux/build.go
claire1618: I could, but I figured it would be good because I'm pretty sure every sRaw includes name, value, images, and created_at? But if that's still a concern I can add that

plyr4: yeah i suppose youre right. but worst case scenario... the worker will crash :(

claire1618: Oof ok yeah will add!

executor/linux/build.go
```diff
@@ -520,6 +535,66 @@ func (c *client) ExecBuild(ctx context.Context) error {
...
+			}
+
+			_log.SetData([]byte(""))
+
+			// ignore pulling secrets coming from plugins
+			if !secret.Origin.Empty() {
+				continue
+			}
+
+			c.Logger.Infof("pulling secret %s", secret.Name)
```
plyr4, line 552: thoughts on moving this `Infof` down below the `continue`? the log makes me think its pulling everything

timhuynh94 (dismissed):

executor/linux/step.go
```diff
@@ -316,7 +317,7 @@ func (c *client) StreamStep(ctx context.Context, ctn *pipeline.Container) error
 					// send API call to append the logs for the step
 					//
 					// https://pkg.go.dev/github.com/go-vela/sdk-go/vela?tab=doc#LogStep.UpdateStep
-					_, err = c.Vela.Log.UpdateStep(c.repo.GetOrg(), c.repo.GetName(), c.build.GetNumber(), ctn.Number, _log)
+					_, err := c.Vela.Log.UpdateStep(c.repo.GetOrg(), c.repo.GetName(), c.build.GetNumber(), ctn.Number, _log)
```
wass3rw3rk, line 320: was this a bug?

wass3rw3rk (commented):
lgtm after removing the `//nolint:...` the linter is complaining about.

we will have to add one for `ExecBuild` function though to quiet down the linter.

executor/linux/step.go
claire1618: I believe so

wass3rw3rk (dismissed):

wass3r (approved):

cognifloyd:
This will only work with the docker runtime. The kubernetes runtime cannot change the container Environment after the build starts. Is there somewhere we can add something to ask the runtime if it supports editing the env?

ecrupper (approved):

wass3rw3rk:
@cognifloyd good callout. so basically, this feature would not be available in k8s. would it just error out? we can create a story for handling that nicer for k8s.

cognifloyd:
I don't think it will error, you will just not see the secret for your steps like you expect--the var will simply be missing.

wass3r:
So, for user facing messaging maybe we print something in the logs to inform the user that lazy loading isn't supported in the environment they are attempting to use it in along with a list of affected secret names? We'll probably want to call that out in docs also. What do you think?

plyr4:
> This will only work with the docker runtime. 
@cognifloyd I tested it on my `kind` cluster locally, and it worked okay. but i am not super familiar with the limitation you described.

plyr4:
hm... i just went back and re-ran those tests and it worked once, then not, im not sure what is inconsistent, but looking at pod events i think it might have been killed and restarted
```
  Normal  Scheduled  2m2s                 default-scheduler  Successfully assigned default/plyr4-heyvela-8 to kind-control-plane
  Normal  Created    2m1s                 kubelet            Created container step-plyr4-heyvela-8-read-fake-secrets-again-again
Normal  Killing    49s                  kubelet            Container step-plyr4-heyvela-8-read-fake-secrets-again-again definition changed, will be restarted
```

but yeah i take it back, doesnt work on k8s

wass3r (approved):

ecrupper (approved):

wass3r (approved):

PR #527, opened 2023-10-25
Fix(worker): on_start failures resulted in success step

timhuynh94:
Closes [#850](https://github.com/go-vela/community/issues/850)

### DIFF
--- a/internal/step/snapshot.go
+++ b/internal/step/snapshot.go
@@ -16,8 +16,10 @@ import (
 // Snapshot creates a moment in time record of the
 // step and attempts to upload it to the server.
 func Snapshot(ctn *pipeline.Container, b *library.Build, c *vela.Client, l *logrus.Entry, r *library.Repo, s *library.Step) {
-	// check if the build is not in a canceled status
-	if !strings.EqualFold(s.GetStatus(), constants.StatusCanceled) {
+	// check if the build is not in a canceled status or error status
+	logrus.Debugf("Snapshot s: %s %s", s.GetName(), s.GetStatus())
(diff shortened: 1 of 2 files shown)


claire1618 (approved):

cognifloyd (approved):

internal/step/snapshot.go
```diff
@@ -16,8 +16,10 @@ import (
 // Snapshot creates a moment in time record of the
 // step and attempts to upload it to the server.
 func Snapshot(ctn *pipeline.Container, b *library.Build, c *vela.Client, l *logrus.Entry, r *library.Repo, s *library.Step) {
-	// check if the build is not in a canceled status
-	if !strings.EqualFold(s.GetStatus(), constants.StatusCanceled) {
+	// check if the build is not in a canceled status or error status
+	logrus.Debugf("Snapshot s: %s %s", s.GetName(), s.GetStatus())
+	if !strings.EqualFold(s.GetStatus(), constants.StatusCanceled) &&
+		!strings.EqualFold(s.GetStatus(), constants.StatusError) {
```
wass3r, line 22: Was this an existing oversight?

wass3r (commented):
Looks good - just a quick question on the Snapshot change. Seems like it would affect other things but I haven't been in this part of the code in a while.

internal/step/snapshot.go
timhuynh94: Yes. I noticed this when the container was in error status and got set to success status here.

plyr4 (approved):

PR #530, opened 2023-10-31
feat: specify which privileged image is being pulled for denied build

claire1618:
When the `VELA_EXECUTOR_ENFORCE_TRUSTED_REPOS` has been enabled, and an untrusted repo has been denied a build, the resulting error is

`error:build denied, repo must be trusted in order to run privileged images`

It would be convenient if the error specified which image is privileged.

The changes in this PR capture the last privileged image and returns that in the error message.

### DIFF
--- a/cmd/vela-worker/exec.go
+++ b/cmd/vela-worker/exec.go
@@ -268,6 +268,14 @@ func (w *Worker) exec(index int, config *library.Worker) error {
 		return nil
 	}
 
+	logger.Info("assembling build")
+	// assemble the build with the executor
+	err = _executor.AssembleBuild(timeoutCtx)
+	if err != nil {
+		logger.Errorf("unable to assemble build: %v", err)
+		return nil
+	}
+
 	// add StreamBuild goroutine to WaitGroup
 	wg.Add(1)
 
@@ -281,14 +289,6 @@ func (w *Worker) exec(index int, config *library.Worker) error {
 			logger.Errorf("unable to stream build logs: %v", err)
 		}
 	}()
-
-	logger.Info("assembling build")
-	// assemble the build with the executor
-	err = _executor.AssembleBuild(timeoutCtx)
-	if err != nil {
-		logger.Errorf("unable to assemble build: %v", err)
-		return nil
-	}
 
 	logger.Info("executing build")
 	// execute the build with the executor
--- a/cmd/vela-worker/exec.go
+++ b/cmd/vela-worker/exec.go
@@ -268,14 +268,6 @@ func (w *Worker) exec(index int, config *library.Worker) error {
 		return nil
 	}
 
-	logger.Info("assembling build")
-	// assemble the build with the executor
-	err = _executor.AssembleBuild(timeoutCtx)
-	if err != nil {
-		logger.Errorf("unable to assemble build: %v", err)
-		return nil
-	}
-
 	// add StreamBuild goroutine to WaitGroup
 	wg.Add(1)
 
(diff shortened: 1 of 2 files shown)


timhuynh94 (dismissed):

executor/linux/build.go
```diff
@@ -370,6 +370,7 @@ func (c *client) AssembleBuild(ctx context.Context) error {
 
 		// assume no privileged images are in use
 		containsPrivilegedImages := false
+		privImage := ""
```
plyr4, line 373: thoughts on making this a slice, so in the scenario multiple images get denied we can log them all with `%v`?

cmd/vela-worker/exec.go
```diff
@@ -268,6 +268,14 @@ func (w *Worker) exec(index int, config *library.Worker) error {
 		return nil
 	}
 
+	logger.Info("assembling build")
+	// assemble the build with the executor
+	err = _executor.AssembleBuild(timeoutCtx)
```
plyr4, line 273: whats the reason for moving Assemble below Stream?

claire1618: I would say that Easton could probably add more to this, but when testing this it kept waiting 5 minutes before it would actually run the build and swapping this code around fixed that.

executor/linux/build.go
claire1618: oh yeah that's a good idea! done!

cmd/vela-worker/exec.go
plyr4: hm interesting... would that be a separate issue? or is it related to the privileged images logging?

claire1618: I made it a separate issue!

executor/linux/build.go
```diff
@@ -370,6 +370,7 @@ func (c *client) AssembleBuild(ctx context.Context) error {
 
 		// assume no privileged images are in use
 		containsPrivilegedImages := false
+		privImage := []string{""}
```
plyr4, line 373:
```suggestion
		privImages := []string{}
```

executor/linux/build.go
```diff
@@ -410,6 +411,7 @@ func (c *client) AssembleBuild(ctx context.Context) error {
 				if privileged {
 					// pipeline contains at least one privileged image
 					containsPrivilegedImages = privileged
+					privImage = append(privImage, container.Image)
```
plyr4, line 414:
```suggestion
					privImages = append(privImages, container.Image)
```

executor/linux/build.go
```diff
@@ -421,8 +423,8 @@ func (c *client) AssembleBuild(ctx context.Context) error {
 
 		// ensure pipelines containing privileged images are only permitted to run by trusted repos
 		if (containsPrivilegedImages) && !(c.repo != nil && c.repo.GetTrusted()) {
-			// update error
-			c.err = fmt.Errorf("unable to assemble build. pipeline contains privileged images and repo is not trusted")
+			// update error including privileged image
+			c.err = fmt.Errorf("unable to assemble build. pipeline contains privileged images and repo is not trusted. privileged image: %v", privImage)
```
plyr4, line 427:
```suggestion
			c.err = fmt.Errorf("unable to assemble build. pipeline contains privileged images and repo is not trusted. privileged image: %v", privImages)
```

plyr4 (approved):
thanks for the tweaks, looks great : )

timhuynh94 (approved):

PR #532, opened 2023-11-02
enhance: assemble build code moved

claire1618:
This fix goes along with [](https://github.com/go-vela/worker/pull/530).

### DIFF
--- a/cmd/vela-worker/exec.go
+++ b/cmd/vela-worker/exec.go
@@ -268,6 +268,14 @@ func (w *Worker) exec(index int, config *library.Worker) error {
 		return nil
 	}
 
+	logger.Info("assembling build")
+	// assemble the build with the executor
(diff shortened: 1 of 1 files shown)


wass3rw3rk:
can you update the description and elaborate on the need for this change?

cognifloyd (changes requested):
Moving this about Streaming does not make sense. Can't AssembleBuild inject logs into some steps? So, streaming needs to (be able to) start before running AssembleBuild.

Are you running into an issue with missing logs or something? What issue does this fix? Or how is this an enhancement?

cognifloyd:
Looks like there is some discussion here:
https://github.com/go-vela/worker/pull/530#discussion_r1378830034

I think moving the AssembleBuild call is just hiding whatever the underlying issue is. Can we find/fix the actual root cause?

PR #539, opened 2023-11-12
fix(secrets): isolate lazy secret loading

wass3r:
this isolates operations for lazy loaded secrets and makes sure the secrets go through the same process as regular secrets. the current implementation had the potential to inadvertently result in unexpected secret values (even for regular secrets), in particular for multiline secrets and/or secrets with escaped newlines.

this also moves the `escapeNewlineSecrets()` on `c.Secrets` to be called once, instead of for however many steps (and services) a pipeline had. it was calling it over and over for the same secrets. this is not necessary for this fix, it's just excessive. if there's concern, i'm happy to remove from this PR.

### DIFF
--- a/executor/linux/build.go
+++ b/executor/linux/build.go
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/go-vela/types/constants"
+	"github.com/go-vela/types/library"
 	"github.com/go-vela/worker/internal/build"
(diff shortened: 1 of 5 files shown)


ecrupper (approved):
Thanks for going through all that — LGTM

timhuynh94 (approved):

KellyMerrick (approved):
lgtm

PR #544, opened 2023-11-22
fix(exec): move `secret.exec` to `ExecBuild` and assemble prior to streaming

ecrupper:
Prior to this change, any error resulting from `AssembleBuild` would cause the worker to enter a 5 minute sleep waiting for the streaming timeout. These errors come primarily in the form of docker images unknown to the registry.

While this may not be a noticeable problem in a large ecosystem, taking workers out of the queue pool for so long could lead to a major increase in queue size if enough workers are stuck waiting.

I figured these changes would be the least invasive way of remedying the above problem. My reasoning is this:

- `AssembleBuild` does not rely on log streaming in any capacity apart from the very last call to `secret.exec`. 
- The `secret.exec` function is hard-coded to report all its logs to the `init` step via log streaming, which means it can safely be removed from `AssembleBuild` without losing its destination for logs.

There was some discussion on this [here](https://github.com/go-vela/worker/pull/532#pullrequestreview-1711077580), where @cognifloyd mentioned that `AssembleBuid` can stream logs from other steps. I did not run into this during my initial testing, but I could be completely missing something.


wass3r (dismissed):

cognifloyd (approved):
I just reviewed the executor and runtime code to validate that `secret.exec` is the only part that depends on log streaming. I agree that that was the only reason to start streaming logs before `AssembleBuild`. I also thought the init step logs went via log streaming, but I see now that they don't; that log gets updated directly.

Moving `secret.exec` from AssembleBuild to ExecBuild feels cleaner and makes the purpose/job of each of these build functions much easier to grok.

Thank you!

executor/linux/build.go
```diff
@@ -469,18 +469,6 @@ func (c *client) AssembleBuild(ctx context.Context) error {
 		return fmt.Errorf("unable to assemble runtime build %s: %w", c.pipeline.ID, c.err)
 	}
 
-	// update the init log with progress
-	//
-	// https://pkg.go.dev/github.com/go-vela/types/library#Log.AppendData
-	_log.AppendData([]byte("> Executing secret images...\n"))
```
cognifloyd, line 475: Should this stay here to keep the output in the init step consistent?

wass3r (approved):

cognifloyd (approved):

claire1618 (approved):

PR #591, opened 2024-07-25
feat: container outputs and dynamic environments

ecrupper:
This PR introduces dynamic environments using a sidecar outputs container.

Proposal ref: https://github.com/go-vela/community/pull/988

## Outputs Container Image

When not set, the build will proceed just as it does today, making this feature completely opt-in and easy to turn off.

The input for `VELA_EXECUTOR_OUTPUTS_IMAGE` will be the image used for the sidecar container that is spun up next to the Vela build. At some point, this could expand to be a programmatic/plugin implementation rather than what it is in this PR — a sleeping container plugged into the volume that is polled after the conclusion of each step.  

In the `docker-compose.yml`, I decided to just go with `alpine:latest`.

## Substitution and Injection relocation

```go
		// perform any substitution on dynamic variables
		err = _step.Substitute()
		if err != nil {
			return err
		}

		c.Logger.Debug("injecting non-substituted secrets")
		// inject no-substitution secrets for container
		err = injectSecrets(_step, c.NoSubSecrets)
		if err != nil {
			return err
		}
```

This block has moved from `CreateStep` to `ExecBuild` for the Docker runtime and will now occur _after_ polling the outputs container. This will allow for dynamic environments such as this:

```yaml
  - name: write to outputs
    image: alpine:latest
    commands:
      - echo "GREETING=hello" >> /vela/outputs/.env
      - echo "SECRET=whisper" >> /vela/outputs/masked.env
      - echo "IMAGE=ubuntu" >> /vela/outputs/.env
  
  - name: read from outputs
    image: ${IMAGE}       # resolves to `ubuntu`
    pull: on_start
    commands:
      - echo $GREETING  # prints hello
      - echo $SECRET      # prints ***
```

## Updates to privileged image checking

To protect against dynamic invocations of privileged images using this new feature, parsing images to determine whether they should run will now be at runtime. The logic has moved from `AssembleBuild` to `ExecStep`/`ExecService`. 

Unfortunately this means that builds will run up until that step is set to execute before denying it, which is a slightly worse UX but I think the tradeoff is worth it.


executor/linux/outputs.go
```diff
@@ -0,0 +1,158 @@
...
+	return nil
+}
+
+// exec runs a secret plugins for a pipeline.
+func (o *outputSvc) exec(ctx context.Context, _outputs *pipeline.Container) error {
+	// exit if outputs container has not been configured
+	if len(_outputs.Image) == 0 {
+		return nil
+	}
+
```
plyr4, line 90: need le image
```
err := o.client.Runtime.CreateImage(ctx, _outputs)
```

wass3r (commented):
was doing some functional testing this morning and was not able to get it working with stages. that's not an intentional limitation, right?

executor/linux/outputs.go
```diff
@@ -0,0 +1,158 @@
...
+// toMap is a helper function that turns raw docker exec output bytes into a map
+// by splitting on carriage returns + newlines and once more on `=`.
+func toMap(input []byte) map[string]string {
+	// carriage returns are included in the split because the exec config `TTY` value is set to true
+	lines := strings.Split(string(input), "\r\n")
+
+	m := make(map[string]string)
+
+	for _, line := range lines {
+		parts := strings.Split(line, "=")
```
wass3rw3rk, line 146: should we use SplitN here, since the value could contain `=`?

wass3rw3rk (commented):
just a question/suggestion

executor/linux/outputs_test.go
```diff
@@ -487,6 +487,15 @@ func TestLinux_Outputs_toMap(t *testing.T) {
 				"TEST": "hello world",
 			},
 		},
+		{
+			name:    "multiple equals",
+			runtime: constants.DriverDocker,
+			input:   []byte("FOO=bar\r\nEQUATION=e=mc^2\r\n"),
+			wantMap: map[string]string{
+				"FOO":      "bar",
+				"EQUATION": "e=mc^2",
```
wass3rw3rk, line 496: 🤣

wass3rw3rk (dismissed):

wass3rw3rk:
friendly reminder to also adds an accompanying docs PR :D

executor/linux/outputs.go
```diff
@@ -0,0 +1,158 @@
...
+			if !strings.Contains(parts[1], "\\\\n") {
+				s = strings.Replace(parts[1], "\\n", "\\\n", -1)
+			}
+
+			m[parts[0]] = s
+		}
+	}
+
+	return m
+}
```
wass3r, line 158: i think it's ok to be less forgiving than something like https://github.com/hashicorp/go-envparse/blob/main/envparse_test.go, but i do wonder if we should account for a little more?

also, line 150, we should change to use `.ReplaceAll` instead

wass3rw3rk (approved):

executor/linux/opts.go
```diff
@@ -185,6 +185,17 @@ func WithVersion(version string) Opt {
 	}
 }
 
+func WithOutputCtn(ctn *pipeline.Container) Opt {
```
plyr4, line 188:
```suggestion
// WithOutputCtn sets the outputs container in the executor client for Linux.
func WithOutputCtn(ctn *pipeline.Container) Opt {
```

executor/linux/outputs.go
```diff
@@ -0,0 +1,150 @@
...
+
+// destroy cleans up outputs container after execution.
+func (o *outputSvc) destroy(ctx context.Context, ctn *pipeline.Container) error {
+	// exit if outputs container has not been configured
+	if len(ctn.Image) == 0 {
+		return nil
+	}
+
+	// update engine logger with secret metadata
```
plyr4, line 63:
```suggestion
	// update engine logger
```

plyr4 (approved):

plyr4:
love it.
any thoughts on letting platform admins adjust the amount of resources allocated to the outputs container? (not blocking)

PR #596, opened 2024-09-18
enhance(operate): check in loop more responsive to termination signal

ecrupper:
If the worker has no build running, terminating the worker should be far more responsive than letting the check in interval expire. This change introduces a 5 second ticker that checks if the context is done or if the interval has been reached for check in.


wass3rw3rk (approved):
lgtm - ty

plyr4 (approved):

PR #615, opened 2024-11-09
fix(secret): update env values for secret origin ctn

ecrupper:
We've added a couple features to the worker (OIDC and Outputs), yet we didn't update the secret container config to account for these. 

These are both very important when considering masked outputs and Vault OIDC policies.


wass3r (approved):

plyr4 (approved):
nice

PR #625, opened 2025-02-21
refactor(skip): use new skip logic for stages

ecrupper:
Follow up from https://github.com/go-vela/server/pull/1255 in the server code.

A couple callouts:

- The `Parallel` field is going away, as it is no longer necessary for the `container.Execute` function. Stage status is now tracked and applied based on the stage's independency.
- Outputs application got shifted up a bit in the code so that the `Skip` call happens after outputs have been added to the env, which will allow for dynamic rulesets with `eval` (recently introduced)


wass3rw3rk (approved):
wouldn't mind a bit more juicy info in the description

plyr4 (approved):
nice, stage status could simplify and speed up the graph endpoint too if we exposed it in the api somehow.
anyways LGTM

PR #626, opened 2025-02-21
fix: update local executor code to include outputs

ecrupper:
Updating local exec capabilities

### DIFF
--- a/executor/local/build.go
+++ b/executor/local/build.go
@@ -237,6 +237,11 @@ func (c *client) AssembleBuild(ctx context.Context) error {
 	// output a new line for readability to stdout
 	fmt.Fprintln(c.stdout, "")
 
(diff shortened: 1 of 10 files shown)


KellyMerrick (approved):

executor/local/build.go
```diff
@@ -294,6 +308,22 @@ func (c *client) ExecBuild(ctx context.Context) error {
 			return fmt.Errorf("unable to plan step: %w", c.err)
 		}
 
+		// poll outputs
+		opEnv, maskEnv, c.err = c.outputs.poll(ctx, c.OutputCtn)
+		if c.err != nil {
+			return fmt.Errorf("unable to exec outputs container: %w", c.err)
```
wass3rw3rk, line 314:
```suggestion
			return fmt.Errorf("unable to poll outputs container: %w", c.err)
```

PR #654, opened 2025-08-09
feat!: add container resource limits and build isolation tracking

delta592:
This PR shouldn't be this crazy, but @wass3r absolutely required I fix many years of reviewdog failures before @wass3r would even considering a review and merge. @plyr4 @ecrupper i don't mind, but it ballooned my original changes, so this review might take a while to review.

My actual changes are in commit https://github.com/go-vela/worker/pull/654/commits/58cc3f0801fd189699be03e62875e2dc09230064

here's hoping @wass3r doesn't need codecov to review. if so, i can attempt to fix those as well.


delta592:
Thanks for the context and guidance! Now I understand more what needs to be fixed for only related items in CI.

I did open a separate PR #655 for just the code changes and those respective golangci-lint and tests. I can close this PR in favor of the other.

PR #659, opened 2025-08-29
chore(lint): update worker codebase

ecrupper:
A few Docker function updates to keep an eye on. Should be no functional changes


executor/linux/build.go
```diff
@@ -290,8 +290,7 @@ func (c *client) AssembleBuild(ctx context.Context) error {
 	for _, s := range c.pipeline.Stages {
 		// TODO: remove hardcoded reference
 		//
-		//nolint:goconst // ignore making a constant for now
-		if s.Name == "init" {
+		if s.Name == constants.InitName {
```
wass3r, line 293: we can drop the comment here (and in many other places) for this now

runtime/kubernetes/pod_tracker.go
```diff
@@ -183,8 +183,8 @@ func newPodTracker(log *logrus.Entry, clientset kubernetes.Interface, pod *v1.Po
 		return nil, fmt.Errorf("newPodTracker expected a pod, got nil")
 	}
 
-	trackedPod := pod.ObjectMeta.Namespace + "/" + pod.ObjectMeta.Name
-	if pod.ObjectMeta.Name == "" || pod.ObjectMeta.Namespace == "" {
+	trackedPod := pod.Namespace + "/" + pod.Name
+	if pod.Name == "" || pod.Namespace == "" {
 		return nil, fmt.Errorf("newPodTracker expects pod to have Name and Namespace, got %s", trackedPod)
```
wass3r, line 188: i know it's not related to your change, but error message doesn't seem to reflect the condition we're checking?

wass3rw3rk (approved):

PR #700, opened 2026-04-15
fix(artifacts): stream log to UI

timhuynh94:
Screenshot 2026-04-16 at 8 54 27 AM

### DIFF
--- a/executor/linux/outputs.go
+++ b/executor/linux/outputs.go
@@ -225,19 +225,75 @@ func (o *outputSvc) pollFiles(ctx context.Context, ctn *pipeline.Container, _ste
 	// https://pkg.go.dev/github.com/sirupsen/logrus#Entry.WithField
 	logger := o.client.Logger.WithField("artifact-outputs", ctn.Name)
 
+	// load the step log so artifact messages are streamed to the UI.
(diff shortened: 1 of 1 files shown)


executor/linux/outputs.go
```diff
@@ -225,20 +225,76 @@ func (o *outputSvc) pollFiles(ctx context.Context, ctn *pipeline.Container, _ste
 	// https://pkg.go.dev/github.com/sirupsen/logrus#Entry.WithField
 	logger := o.client.Logger.WithField("artifact-outputs", ctn.Name)
 
+	// load the step log so artifact messages are streamed to the UI.
+	// Wait briefly to allow the StreamStep goroutine to finish its final
+	// log upload so that we do not race with it. Then re-fetch the log
+	// from the server to get the authoritative copy that includes all
+	// container output — this prevents artifact messages from overwriting
+	// the step's own logs.
+	time.Sleep(2 * time.Second)
```
wass3rw3rk, line 234: really not a fan of putting an arbitrary wait time here. is there some kind of event to hook into?

wass3rw3rk (changes requested):

